PR #18, opened 2021-04-11
Rework on MacosApp

bdlukaa:
Reworked MacosApp to use CupertinoApp as a backend. This has several improvements:

- Apple-styled scrolling
- Using the San Francisco font by default

I also added documentation for all the properties

### DIFF
--- a/lib/src/macos_app.dart
+++ b/lib/src/macos_app.dart
@@ -1,25 +1,58 @@
 import 'package:macos_ui/macos_ui.dart';
-import 'package:flutter/material.dart' as m;
+import 'package:flutter/cupertino.dart' as c;
 
-/// Defines an application that uses macOS design
+/// An application that uses macOS design.
 ///
-/// todo: Documentation
+/// A convenience widget that wraps a number of widgets that are commonly
+/// required for an iOS-design targeting application. It builds upon a
+/// [WidgetsApp] by iOS specific defaulting such as fonts and scrolling
+/// physics.
+///
+/// The [MacosApp] configures the top-level [Navigator] to search for routes
+/// in the following order:
+///
+///  1. For the `/` route, the [home] property, if non-null, is used.
+///
+///  2. Otherwise, the [routes] table is used, if it has an entry for the route.
+///
+///  3. Otherwise, [onGenerateRoute] is called, if provided. It should return a
+///     non-null value for any _valid_ route not handled by [home] and [routes].
+///
+///  4. Finally if all else fails [onUnknownRoute] is called.
+///
+/// If [home], [routes], [onGenerateRoute], and [onUnknownRoute] are all null,
+/// and [builder] is not null, then no [Navigator] is created.
+///
+/// This widget also configures the observer of the top-level [Navigator] (if
+/// any) to perform [Hero] animations.
 class MacosApp extends StatefulWidget {
+  /// Creates a MacosApp.
+  ///
+  /// At least one of [home], [routes], [onGenerateRoute], or [builder] must be
+  /// non-null. If only [routes] is given, it must include an entry for the
+  /// [Navigator.defaultRouteName] (`/`), since that is the route used when the
+  /// application is launched with an intent that specifies an otherwise
+  /// unsupported route.
+  ///
+  /// This class creates an instance of [WidgetsApp].
+  ///
+  /// The boolean arguments, [routes], and [navigatorObservers], must not be null.
   const MacosApp({
     Key? key,
     this.navigatorKey,
+    this.home,
+    Map<String, Widget Function(BuildContext)> this.routes =
+        const <String, WidgetBuilder>{},
+    this.initialRoute,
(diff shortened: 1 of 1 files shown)


bdlukaa:
The documentation was partially copied from `CupertinoApp`

GroovinChip:
Wow, thanks so much for the PR!! I will review soon.

GroovinChip:
> The documentation was partially copied from `CupertinoApp`

I was planning on doing that myself so no worries 😅

lib/src/macos_app.dart
```diff
@@ -1,25 +1,58 @@
 import 'package:macos_ui/macos_ui.dart';
-import 'package:flutter/material.dart' as m;
+import 'package:flutter/cupertino.dart' as c;
 
-/// Defines an application that uses macOS design
+/// An application that uses macOS design.
 ///
-/// todo: Documentation
+/// A convenience widget that wraps a number of widgets that are commonly
```
GroovinChip, line 6: Change iOS references to macOS

lib/src/macos_app.dart
```diff
@@ -65,100 +95,196 @@ class MacosApp extends StatefulWidget {
...
         super(key: key);
 
-  final Map<Type, Action<Intent>>? actions;
-
-  final BackButtonDispatcher? backButtonDispatcher;
-
-  final TransitionBuilder? builder;
-
-  final bool checkerboardOffscreenLayers;
+  /// {@macro flutter.widgets.widgetsApp.navigatorKey}
```
GroovinChip, line 116: Run dartdoc and see how the macros render - I'm not sure if they will show without a dartdoc options yaml file. If the macros do not render, either remove them or configure dartdoc to render them.

lib/src/macos_app.dart
```diff
@@ -65,100 +95,196 @@ class MacosApp extends StatefulWidget {
...
+  final GlobalKey<NavigatorState>? navigatorKey;
```
GroovinChip, line 117: Sort class members alphabetically, per code_metrics analysis

lib/src/macos_app.dart
```diff
@@ -191,23 +317,21 @@ class _MacosAppState extends State<MacosApp> {
...
+          color: theme.typography?.body?.color,
         ),
+        child: child!,
       ),
     );
   }
 
   Widget _buildApp(BuildContext context) {
-    final defaultColor = widget.primaryColor ?? CupertinoColors.systemBlue;
+    final defaultColor = widget.color ?? CupertinoColors.systemBlue;
```
GroovinChip: While `WidgetsApp` and `CupertinoApp` do call this property `color`, I feel `primaryColor` is more accurate. That said, an even more accurate name would probably be `systemColor`, since the documentation for this property defines it as being the primary color used by the OS for the app.

GroovinChip (changes requested):

lib/src/macos_app.dart
bdlukaa: But nobody uses dartdoc. When reading these documentations, developers usually use vscode (which works) or api.pub.dev (which also works): https://api.flutter.dev/flutter/cupertino/CupertinoApp/navigatorKey.html

lib/src/macos_app.dart
bdlukaa: `primaryColor` is definitely better than `systemColor`, because the dev may think `sytemColor` is the color used by the system, not the app.

GroovinChip: Fair enough, let's go with `primaryColor`

lib/src/macos_app.dart
GroovinChip: pub.dev uses dartdoc to generate documentation for packages published, and I would certainly like to see this published to pub someday 😅

bdlukaa: This makes no sense. See https://pub.dev/packages/dartdoc#macros

GroovinChip: I ran dartdoc myself, just to see, and it looks like it doesn't matter after all because since we're exporting material, the material docs get built 😅

Screen Shot 2021-04-11 at 9 03 09 PM

lib/src/macos_app.dart
bdlukaa: Why? From my point of view this is useless. Just extra unecessary work

GroovinChip: Alright, no worries

lib/src/macos_app.dart
bdlukaa: Nice!

GroovinChip: I mean its not really desired behavior, but I can look into that another time 😅

bdlukaa:
Fixed the changes

GroovinChip (approved):

GroovinChip:
LGTM, thank you so much for the PR!!

PR #29, opened 2021-04-16
Theme refactor

### LINKED ISSUE #25: Rename Style to MacosThemeData

GroovinChip:
* Major theme refactor that more closely resembles flutter/material and flutter/cupertino
  * The `Style` class is now `MacosThemeData`
  * `MacosTheme` is now a `StatelessWidget` that returns a private `_InheritedMacosTheme`.  The static `MacosTheme.of(context)` is now defined here.
  * `MacosApp` now takes a `theme` and `darkTheme` rather than `style` and `darkStyle`.  Additionally, there are minor changes to the way `MacosApp` is built that more closely resemble how `MaterialApp` is built. 

Inspired by:
* material/theme.dart
* material/theme_data.dart
* material/app.dart
* cupertino/theme.dart

Closes #25


lib/src/buttons/switch.dart
```diff
@@ -65,7 +65,7 @@ class Switch extends StatelessWidget {
       onChanged: onChanged,
       dragStartBehavior: dragStartBehavior,
       activeColor: activeColor ??
-          context.maybeStyle?.primaryColor ??
+          context.macosTheme.primaryColor ??
```
bdlukaa: This should be `context.macosTheme?.primaryColor` because we are not calling `debugCheckHasMacosTheme`

GroovinChip: Good catch

bdlukaa: Or add the check `debugCheckHasMacosTheme` on every `build` method at each widget

GroovinChip: Don't think it's necessary to call it in every widget's `build` method, doing it at the `MacosApp` level should be fine I think

bdlukaa:
I think the last commit should be reverted. `primaryColor` can be null. The think is that `debugCheckHasMacosTheme` should be called at the top of the build function, just like in `Scaffold`.

GroovinChip:
@bdlukaa  see https://github.com/GroovinChip/macos_ui/blob/theme_refactor/lib/src/styles/macos_theme_data.dart#L27 there will always be a `primaryColor` now.

lib/src/styles/macos_theme_data.dart
```diff
@@ -0,0 +1,100 @@
...
+    Brightness? brightness,
+    Color? primaryColor,
+    Color? accentColor,
+    Curve? animationCurve,
+    Duration? mediumAnimationDuration,
+    Typography? typography,
+  }) {
+    final Brightness _brightness = brightness ?? Brightness.light;
+    final bool isDark = _brightness == Brightness.dark;
+    primaryColor ??= isDark
```
bdlukaa, line 27: You are not considering when the device is in high contrast mode. `CupertinoDynamicColor` have all the contrast options

GroovinChip: This is true. I was intending to add it later.

bdlukaa: Actually, I think this is not necessary, because the color handling should be up to the widget (`CupertinoDynamicColors.resolve(color, context)`). So, you should not be handling if the brightness is dark or not, because `CupertinoDynamicColor` already have the dark color variation. If you want to add dark mode to a color, make it a CupertinoDynamicColor and set the dark property

GroovinChip: I'm a little confused, would you mind clarifying what you think should/shouldn't be changed?

bdlukaa: You're checking if the current theme is dark to set a color, but it shouldn't be done this way. `CupertinoDynamicColor` have all the color variants. If it needs to have a dark theme variation, it should be set in `CupertinoDynamicColor`.

GroovinChip: Ah, gotcha. I remember seeing that flutter/cupertino has a way to handle that. I'll look into it more. That said, I'm inclined to say that not handling high contrast shouldn't be a merge-blocker at this point.

lesliearkorful (approved):

PR #30, opened 2021-04-16
Implemented Level indicator

### LINKED ISSUE #38: Add `LevelIndicator` widget

bdlukaa:
* Continuous capacity indicator 
* Discrete capacity indicator 
* Rating indicator 
* Relevance indicator


lib/src/indicators/rating_indicator.dart
```diff
@@ -0,0 +1,87 @@
...
+      onPanUpdate: (event) => _handleUpdate(event.localPosition),
+      onPanDown: (event) => _handleUpdate(event.localPosition),
+      child: Row(
+        mainAxisSize: MainAxisSize.min,
+        children: List.generate(amount, (index) {
+          final rated = value > index;
+          return Icon(
+            rated ? ratedIcon : unratedIcon,
+            color: iconColor ??
+                context.maybeStyle?.primaryColor ??
```
GroovinChip: There is no more `maybeStyle`

GroovinChip (changes requested):
Ensure `dartfmt` and optimize imports have been run on all changed files in addition to other requested changes.

lib/src/indicators/rating_indicator.dart
bdlukaa: Done

GroovinChip (changes requested):
Track colors for CapacityIndicators and RelevanceIndicator are not correct for light theme
Screen Shot 2021-04-18 at 10 39 33 AM

bdlukaa:
@GroovinChip This is actually an issue with the `MacosTheme`. `CupertinoDynamicColor.resolveFrom` uses `CupertinoTheme.maybeBrightnessOf(context)`, that uses `MediaQuery.maybeOf(context)?.platformBrightness` for color fallback. A `CupertinoTheme` should be provided by the `MacosTheme` with the brightness, otherwise all the widgets that uses `CupertinoDynamicColor.resolve()` will use the device's brightness instead of the one provided in `MacosTheme`.

GroovinChip:
> @GroovinChip This is actually an issue with the `MacosTheme`. `CupertinoDynamicColor.resolveFrom` uses `CupertinoTheme.maybeBrightnessOf(context)`, that uses `MediaQuery.maybeOf(context)?.platformBrightness` for color fallback. A `CupertinoTheme` should be provided by the `MacosTheme` with the brightness, otherwise all the widgets that uses `CupertinoDynamicColor.resolve()` will use the device's brightness instead of the one provided in `MacosTheme`.

@bdlukaa  I think this will be resolved by #43 

Edit: Actually no. This is a step forward, but not the complete solution. I don't think a `CupertinoTheme` should be added to `MacosTheme`. Rather, I think that we should have an extension on `CupertinoDynamicColor` that calls the new `MacosTheme.maybeBrightnessOf(context)` and handles things accordingly.

GroovinChip (approved):
LGTM

PR #50, opened 2021-04-23
Fixed #49

### LINKED ISSUE #49: Level Indicator not working as expected
As stated in the description provided by apple the Level Indicator should implement user interaction. Clicking on a bar or sliding over the bar should adjust the value.

Screenshot 2021-04-23 at 22 24 38

However, as you can see in the following GIF clicking on a bar does not correspond to the correct selected value.

showcase

As far as I can tell the element you tap on should be selected. 
The sliding is fine but the first click of the slide is registered as tap thus the value changes before the sliding applies. This should be avoided

bdlukaa:
The GestureDetector applied to capacity indicator should be added inside of the Container.

Fixes #49

Bumped version to 0.0.8

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [0.0.8]
+* Fixed [#49](https://github.com/GroovinChip/macos_ui/issues/49)
+
 ## [0.0.7]
 * Implemented `Checkbox`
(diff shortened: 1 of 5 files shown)


GroovinChip:
LGTM

GroovinChip (approved):

Issue #26, opened 2021-04-15
☂️ Revamp Scaffold


GroovinChip:
Discussion here: https://github.com/GroovinChip/macos_ui/discussions/15

- [x] Support multiple resizable panels
- [x] Scaffold should show/hide any sidebars/resizable panes at a specified breakpoint
- [x] Scaffold `Titlebar` should support `Toolbar` widgets

GroovinChip:
Another thing - Toolbar and ResizablePane dividers should respect their boundaries, as seen here:
<img width="766" alt="Screen Shot 2021-04-15 at 10 38 30 PM" src="https://user-images.githubusercontent.com/4250470/114963694-50c0be00-9e3b-11eb-9f7a-1c32a2150e08.png">
- [ ] No horizontal toolbar divider for Sidebars

lesliearkorful:
Sidebars have a horizontal divider at the top when they're scrolled.
Even more like a shadow
![image](https://user-images.githubusercontent.com/4087388/115014344-ba73b300-9ea1-11eb-90c6-de36acfebc50.png)

GroovinChip:
Ah, you're correct.

GroovinChip:
@whiplashoo this will play into #80

dbsxdbsx:
I just wonder how to add vertical divider to separate item in toolbar? Now I only found `ToolBarSpacer`.

whiplashoo:
@dbsxdbsx that's not available yet, but definitely in the roadmap. If you want, please open a separate issue to track it.

GroovinChip:
@dbsxdbsx Yes, please feel free to open an issue for this feature.

PR #52, opened 2021-04-26
Scaffold

lesliearkorful:
Updated work on #26 

image

## Pre-launch Checklist


lib/src/layout/resizable_pane.dart
```diff
@@ -0,0 +1,179 @@
...
+/// Indicates the draggable side of the sidebar for resizing
+enum ResizableSide { left, right }
+
+class ResizablePane extends StatefulWidget {
+  /// Creates a widget that can be resized horizontally.
+  ///
+  /// The [builder], [minWidth] and [resizableSide] can not be null.
+  /// The [maxWidth] and the [scaffoldBreakpoint] default to `500.00`.
+  /// [isResizable] defaults to `true`.
+  ResizablePane({
```
bdlukaa: Make the widgets `const`

lib/src/layout/sidebar.dart
```diff
@@ -0,0 +1,69 @@
+import 'package:flutter/widgets.dart';
+import 'package:macos_ui/macos_ui.dart';
+
+class Sidebar {
+  /// Creates a macOS-style side bar at left side of the [Scaffold].
+  ///
+  /// The [builder] and [minWidth] properties are required.
+  /// The sidebar builds with a scrollbar internally.
+  Sidebar({
```
bdlukaa: `const`

lib/src/layout/sidebar.dart
```diff
@@ -0,0 +1,69 @@
...
+  /// /// The minimum width that this sidebar can be resized to.
+  ///
+  /// The [minWidth] should not be more than the [maxWidth].
+  final double minWidth;
+
+  /// Empty space to inscribe inside the title bar. The [child], if any, is
+  /// placed inside this padding.
+  ///
+  /// Defaults to `EdgeInsets.only(top: 52.0)`.
+  final EdgeInsets? padding;
```
bdlukaa: If padding has a default, it shouldn't be nullable. `EdgeInsets.only(top: 52.0)` should be defined by default on the constructor

lib/src/layout/title_bar.dart
```diff
@@ -0,0 +1,47 @@
...
+  final Widget? child;
+
+  /// The decoration to paint behind the [child].
+  final BoxDecoration? decoration;
+
+  /// Empty space to inscribe inside the title bar. The [child], if any, is
+  /// placed inside this padding.
+  ///
+  /// Defaults to `EdgeInsets.all(8)`
+  final EdgeInsets? padding;
```
bdlukaa: here too

bdlukaa (approved):
LGTM

lib/src/layout/resizable_pane.dart
```diff
@@ -0,0 +1,180 @@
...
+
+  @override
+  void dispose() {
+    _scrollController.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    assert(widget.maxWidth! >= widget.minWidth);
```
bdlukaa: If you're gonna keep it this way, make sure to check if `widget.maxWidth` is not null instead of adding `!`.

```dart
if (widget.maxWidth != null)
  assert(widget.maxWidth >= widget.minWidth)
```

This can also be checked on the widget constructor (I recommend that way):

```dart
  const ResizablePane({
    Key? key,
    required this.builder,
    this.decoration,
    this.maxWidth = 500.0,
    required this.minWidth,
    this.isResizable = true,
    required this.resizableSide,
    this.scaffoldBreakpoint = 500,
  }) : assert(maxWidth == null || maxWidth >= minWidth), super(key: key);
```

lib/src/layout/resizable_pane.dart
```diff
@@ -0,0 +1,180 @@
...
+  /// Pass the [scrollController] obtained from this method, to a scrollable
+  /// widget used in this method to work with the internal [Scrollbar].
+  final ScrollableWidgetBuilder builder;
+
+  /// The [BoxDecoration] to paint behind the child in the [builder].
+  final BoxDecoration? decoration;
+
+  /// Specifies if this [ResizablePane] can be resized by dragging the
+  /// resizable side of this widget.
+  final bool? isResizable;
```
bdlukaa: This must not be null. Make it a `bool` instead of `bool?`

lib/src/layout/scaffold.dart
```diff
@@ -1,117 +1,303 @@
+import 'dart:math' as math;
+import 'dart:ui';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/rendering.dart';
 import 'package:macos_ui/macos_ui.dart';
-import 'package:split_view/split_view.dart'; //todo: fork and migrate to nnbd
+import 'package:macos_ui/src/layout/content_area.dart';
```
bdlukaa, line 7: The `import`s below should also be `export`ed

lib/src/layout/scaffold.dart
```diff
@@ -1,117 +1,303 @@
...
+    late Color backgroundColor;
+    late Color sidebarBackgroundColor;
+    Color dividerColor = theme.dividerColor!;
 
-    if (style.brightness == Brightness.light) {
-      bodyColor = backgroundColor ?? CupertinoColors.systemBackground.color;
-      sidebarColor =
-          sidebarBackgroundColor ?? CupertinoColors.tertiarySystemBackground;
-      gripColor = sidebarGripColor ?? CupertinoColors.systemFill;
+    if (!theme.brightness!.isDark) {
```
bdlukaa, line 127: I would recommed to create a new local `CupertinoDynamicColor` to avoid boilerplate code. Just a recommendation, but it's good the way it is

lib/src/layout/scaffold.dart
```diff
@@ -1,117 +1,303 @@
...
+    required this.valueNotifier,
+  }) : super(key: key, child: child);
+
+  final BoxConstraints constraints;
+
+  final double contentAreaWidth;
+
+  final ResizablePaneNotifier valueNotifier;
+
+  static ScaffoldScope of(BuildContext context) {
```
bdlukaa, line 315: There should also be a `maybeOf(BuildContext)` method that returns a nullable `ScaffoldScope?`

bdlukaa (changes requested):

lib/src/layout/scaffold.dart
lesliearkorful: It's exported in `macos_ui.dart`

bdlukaa (approved):

lesliearkorful:
@GroovinChip

lib/src/layout/content_area.dart
```diff
@@ -0,0 +1,39 @@
...
+  const ContentArea({Key? key, @required this.builder}) : super(key: key);
+
+  /// The builder that creates a child to display in this widget, which will
+  /// use the provided [_scrollController] to enable the scrollbar to work.
+  ///
+  /// Pass the [_scrollController] obtained from this method, to a scrollable
+  /// widget used in this method to work with the internal [Scrollbar].
+  final ScrollableWidgetBuilder? builder;
+
+  static final _scrollController = ScrollController();
```
bdlukaa, line 23: I see here you're considering an application to have only one scaffold, when there can be more. I think the `Scrollbar` should be implemented by the dev, not by the `Scaffold` widget

bdlukaa: Also, you mention in the documentation of the property `builder` that the dev can use [_scrollController], but the dev can't because the property is private

lib/src/layout/scaffold.dart
```diff
@@ -1,118 +1,310 @@
...
+    final ScaffoldScope? result =
+        context.dependOnInheritedWidgetOfExactType<ScaffoldScope>();
+    assert(result != null, 'No ScaffoldContraints found in context');
+    return result!;
+  }
+
+  static ScaffoldScope? maybeOf(BuildContext context) {
+    final ScaffoldScope? result =
+        context.dependOnInheritedWidgetOfExactType<ScaffoldScope>();
+    assert(result != null, 'No ScaffoldContraints found in context');
```
bdlukaa: There is no point on asserting here because the method can return null

bdlukaa (changes requested):

lib/src/layout/content_area.dart
lesliearkorful: By default every section of the macOS window is scrollable. I've also heard that flutter considers adding scrollbar to every scrollview by default. I'll track that and update our repo later.
Check here for the scrollController example: https://github.com/GroovinChip/macos_ui/blob/4ce9eb81779acd8e9d971cbe5f189148c5409f8e/example/lib/main.dart#L53

And about having one Scaffold, that's what I've observed so far from the official mac apps. It's up to the developer to customize the UI though. The ScaffoldScope takes the nearest Scaffold, so I think we can have multiple Scaffold.
I'll add the sidebar buttons together with navigation support after this PR lands. Consider this the basic frame.
Scaffold is a long work in progress. We can't wait to factor in everything before merging this. It'll take forever

lib/src/layout/scaffold.dart
```diff
@@ -1,118 +1,331 @@
...
+import 'package:macos_ui/src/layout/sidebar.dart';
+import 'package:macos_ui/src/layout/title_bar.dart';
+
+const _kTitleBarHeight = 52.0;
+
+const _kSmallTitleBarHeight = 30.0;
 
 /// A basic screen-layout widget.
 ///
 /// Provides a [body] for main content and a [sidebar] for secondary content
```
GroovinChip: Body seems to have been replaced by `children`, so this documentation needs to be updated.

bdlukaa (approved):

lib/src/layout/scaffold.dart
```diff
@@ -1,118 +1,331 @@
...
 /// (like navigation buttons). If no [sidebar] is specified, only the [body]
 /// will be shown.
-class Scaffold extends StatelessWidget {
+class Scaffold extends StatefulWidget {
+  /// Creates a macOS window layout.
+  ///
+  /// The [children] can only include one [ContentArea], but can no or multiple [ResizablePane] widgets.
```
GroovinChip: This line needs to be updated, I don't know what "can no or multiple" means

lib/src/layout/scaffold.dart
```diff
@@ -1,118 +1,331 @@
...
+  }) : super(key: key);
+
+  /// Specifies the background color for the Scaffold.
+  ///
+  /// The default colors from the theme would be used if no color is specified.
   final Color? backgroundColor;
 
-  /// Main content area.
-  final Widget body;
+  /// The children to display in the rest of the scaffold, excluding the [Sidebar] and [Titlebar] regions.
```
GroovinChip: Documentation should stay within the 80 character bound

lib/src/layout/scaffold.dart
```diff
@@ -1,118 +1,331 @@
...
+              contentAreaWidth: math.max(_minContentAreaWidth, _remainingWidth),
+            );
+          },
+        );
+      },
     );
   }
 }
+
+class ScaffoldScope extends InheritedWidget {
```
GroovinChip, line 296: This class needs documentation. The class members and functions need docs as well.

lib/src/layout/sidebar.dart
```diff
@@ -0,0 +1,69 @@
+import 'package:flutter/widgets.dart';
+import 'package:macos_ui/macos_ui.dart';
+
+class Sidebar {
+  /// Creates a macOS-style side bar at left side of the [Scaffold].
```
GroovinChip, line 5: This documentation should be above the class declaration

lib/src/layout/scaffold.dart
```diff
@@ -1,118 +1,331 @@
...
+import 'package:flutter/rendering.dart';
 import 'package:macos_ui/macos_ui.dart';
-import 'package:split_view/split_view.dart';
+import 'package:macos_ui/src/layout/content_area.dart';
+import 'package:macos_ui/src/layout/resizable_pane.dart';
+import 'package:macos_ui/src/layout/resizable_pane_notifier.dart';
+import 'package:macos_ui/src/layout/sidebar.dart';
+import 'package:macos_ui/src/layout/title_bar.dart';
+
+const _kTitleBarHeight = 52.0;
```
GroovinChip: Add docs

lib/src/layout/scaffold.dart
```diff
@@ -1,118 +1,331 @@
...
+
+const _kSmallTitleBarHeight = 30.0;
```
GroovinChip: Add docs

lib/src/layout/resizable_pane_notifier.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:flutter/foundation.dart';
+
+class ResizablePaneNotifier extends ValueNotifier<Map<Key, double>> {
+  /// Creates a [ValueNotifier] that wraps this [value] for a [ResizablePane]
```
GroovinChip, line 4: Move this line above the class declaration

lib/src/layout/title_bar.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:flutter/widgets.dart';
+
+enum TitleBarSize { large, small }
+
+class TitleBar {
+  /// Creates a title bar in the [Scaffold].
```
GroovinChip, line 6: Move this documentation above the class declaration

lib/macos_ui.dart
```diff
@@ -10,33 +10,45 @@
 
 library macos_ui;
 
+export 'package:flutter/cupertino.dart'
```
GroovinChip: Remove these, we have `library.dart` for exporting flutter libraries now

GroovinChip (changes requested):
Also update the [version](https://github.com/GroovinChip/macos_ui/blob/029716c1dd2e0094f82dc324e93a6714f57f2e48/pubspec.yaml#L3) in pubspec.yaml

lib/src/layout/scaffold.dart
bdlukaa: I don't think it's necessary to add docs to this one because its private.

lib/src/layout/resizable_pane_notifier.dart
bdlukaa: I don't think this line should be moved up, otherwise the constructor would have no documentation.

lib/src/layout/scaffold.dart
GroovinChip: Fair

lib/src/layout/resizable_pane_notifier.dart
GroovinChip: Class documentation is more important than constructor documentation

vietz-dev: Couldn't we use the dart doc macro here?

https://pub.dev/packages/dartdoc#macros

Example:
https://github.com/flutter/flutter/blob/1d9032c7e1d867f071f2277eb1673e8f9b0274e3/packages/flutter/lib/src/material/button.dart#L133

lib/src/layout/scaffold.dart
bdlukaa: I think she mean "can be one or multiple"

lib/src/layout/scaffold.dart
vietz-dev: Is there a particular reason this variable is private?

The AppBar for example uses `kToolbarHeight` which is public available.
https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/constants.dart#L22

I think it might come in handy to get access to these kinds of parameters as a developer.

lib/src/layout/scaffold.dart
GroovinChip: This can be done later

GroovinChip (approved):

PR #58, opened 2021-05-02
Code Improvements

bdlukaa:
Several code improvements all around the project. This implements the following things:

- Accessibility support for all the widgets
- Diagnostics support for almost all the widgets and classes (devtools)
- `debugCheckHasMacosTheme` now has a better error message
- Improved documentation all around the package
- Replaced `todo` -> `TODO`, because of dart analyzer
- Replaced `CupertinoDynamicColor` -> `DynamicColorX`
- Added `MacosTheme.maybeOf`, which can result in a nullable `MacosThemeData`

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
+## [0.0.9+1]
+* `CapacityIndicator` colors can now be setted on its constructor
+* Accessibility support for most of the widgets
+* Diagnostics Properties (dev tools) for most of the widgets
+
 ## [0.0.9]
-* Implemented 'HelpButton'
-=======
+* Implemented `HelpButton`
 * Fixed [#49](https://github.com/GroovinChip/macos_ui/issues/49)
 
 ## [0.0.7]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## [0.0.9+1]
-* `CapacityIndicator` colors can now be setted on its constructor
+* `CapacityIndicator` colors can now be set on its constructor
 * Accessibility support for most of the widgets
 * Diagnostics Properties (dev tools) for most of the widgets
 
--- a/lib/macos_ui.dart
+++ b/lib/macos_ui.dart
@@ -1,18 +1,14 @@
 library macos_ui;
 
-export 'package:flutter/cupertino.dart'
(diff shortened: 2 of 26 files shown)


lib/macos_ui.dart
```diff
@@ -14,7 +12,6 @@ export 'package:flutter/material.dart'
         FlutterLogo,
         CircleAvatar;
 
-/// todo: package-level docs
```
GroovinChip, line 17: Let's leave this in, because we do need package-level docs

lib/macos_ui.dart
```diff
@@ -1,7 +1,5 @@
 library macos_ui;
 
-export 'package:flutter/cupertino.dart'
```
GroovinChip, line 3: This removal reminds me that we really shouldn't be exporting any non-macos_ui libraries in this file, it messes with the generation of documentation. If we need the convenience of exporting flutter libraries in one file, we should have a separate file specifically for this that is not exported in `lib/macos_ui.dart`. Perhaps a `lib/src/library.dart`.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
+## [0.0.9+1]
+* `CapacityIndicator` colors can now be setted on its constructor
```
GroovinChip: "setted" isn't a word, change to something like "`CapacityIndicator` colors can now be set on its constructor"

GroovinChip (changes requested):
These changes are great, I especially like the accessibility improvements!

lib/macos_ui.dart
GroovinChip: I will tinker with this now

CHANGELOG.md
bdlukaa: Oh sorry. I didn't know that

GroovinChip: No worries 😊

lib/macos_ui.dart
GroovinChip: Made the change, running dartdoc now to see the results

GroovinChip (changes requested):
Looking good. I will submit my own changes on this branch shortly.

GroovinChip (approved):

GroovinChip:
Other changes made:
* Updated `dart_code_metrics` to latest version
* Added a lint rule to avoid relative lib imports

PR #118, opened 2021-05-26
Fix text field

bdlukaa:
Properly aligns the prefix, suffix and clear button to the text. When `maxLines` is null, it gets the padding applied to the `EditableText` and applies to the prefix and suffix. When not, it's aligned to the center using `Row`'s `crossAxisAlignment` property

## Pre-launch Checklist

- [ ] I have run `flutter pub publish --dry-run` and addressed any warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [0.2.4]
+* Fix textfield prefix icon alignment
+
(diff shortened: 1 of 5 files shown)


GroovinChip (dismissed):
Bump version and changelog

GroovinChip (approved):

PR #126, opened 2021-06-02
Colors update

GroovinChip:
This PR updates the `MacosColors` class with dark/light variants of some colors. Additionally, it contains documentation that describes the rather inscrutably-named colors (I blame Apple).

## Pre-launch Checklist

### DIFF
--- a/example/lib/pages/colors_page.dart
+++ b/example/lib/pages/colors_page.dart
@@ -15,62 +15,7 @@ class _ColorsPageState extends State<ColorsPage> {
       children: [
         Wrap(
           spacing: 8.0,
-          runSpacing: 4.0,
-          children: [
-            MacosTooltip(
-              message: 'Label',
-              child: ColorBox(
-                color: MacosColors.labelColor,
-              ),
-            ),
-            MacosTooltip(
-              message: 'Label Dark',
-              child: ColorBox(
-                color: MacosColors.labelColor.darkColor,
-              ),
-            ),
-            MacosTooltip(
-              message: 'Secondary Label',
-              child: ColorBox(
-                color: MacosColors.secondaryLabelColor,
-              ),
-            ),
-            MacosTooltip(
-              message: 'Secondary Label Dark',
-              child: ColorBox(
-                color: MacosColors.secondaryLabelColor.darkColor,
-              ),
-            ),
-            MacosTooltip(
-              message: 'Tertiary Label',
-              child: ColorBox(
-                color: MacosColors.tertiaryLabelColor,
-              ),
-            ),
-            MacosTooltip(
-              message: 'Tertiary Label Dark',
-              child: ColorBox(
-                color: MacosColors.tertiaryLabelColor.darkColor,
-              ),
-            ),
-            MacosTooltip(
-              message: 'Quaternary Label',
-              child: ColorBox(
-                color: MacosColors.quaternaryLabelColor,
-              ),
-            ),
-            MacosTooltip(
-              message: 'Quaternary Label Dark',
-              child: ColorBox(
-                color: MacosColors.quaternaryLabelColor.darkColor,
-              ),
-            ),
-          ],
-        ),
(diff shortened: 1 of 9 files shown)


lib/src/theme/macos_colors.dart
```diff
@@ -17,50 +17,156 @@ class MacosColors {
...
+  /// The background of a window.
   static const windowBackgroundColor = MacosColor(0xff3b373d);
+
+  /// The background behind a document's content.
   static const underPageBackgroundColor = MacosColor(0xff282828);
-  static const controlBackgroundColor = MacosColor(0xff1e1e1e);
+
+  /// The background of a large interface element, such as a browser or table.
+  static const controlBackgroundColor = MacosColor(0xffFFFFFF);
+  static const controlBackgroundColorDark = MacosColor(0xff1E1E1E);
```
bdlukaa: These could be only one:

```dart
static const Color controlBackgroundColor = CupertinoDynamicColor.fromBrightness(
  lightColor: MacosColor(0xffFFFFFF), // Could be MacosColors.white too
  darkColor: MacosColor(0xff1E1E1E),
),
```

bdlukaa (changes requested):
The comment on the two colors applied to many more.

The type for each color should be explicit, just like in `MacosColors.transparent`:
`static const Color white = MacosColor(0xffffffff);` instead of  `static const white = MacosColor(0xffffffff);`

GroovinChip:
> The comment on the two colors applied to many more.
> 
> The type for each color should be explicit, just like in `MacosColors.transparent`:
> `static const Color white = MacosColor(0xffffffff);` instead of `static const white = MacosColor(0xffffffff);`

Any particular reason you think we need to specify the type? `MacosColor` extends `Color`, and Dart's type checker is pretty good at determining the type in cases like this.

lib/src/theme/macos_colors.dart
GroovinChip: Tbh I don't really understand how `fromBrightness` works because we aren't specifying a `Brightness`. I mean I see that it checks Platform Brightness but do we know for a fact that the correct brightness will get returned for a `MacosApp`?

GroovinChip: ~~Yeah I tested this out and it does not work.~~ Well I take that back, it does. Hot reload just wasn't bringing the change out

bdlukaa: > I mean I see that it checks Platform Brightness but do we know for a fact that the correct brightness will get returned for a MacosApp?

`fromBrightness` is only to tell what color to use at the time. To get the correct color, use `MacosDynamicColor.resolve(MacosColor.controlBackgroundColor, context)`

GroovinChip:
@bdlukaa

bdlukaa:
> Any particular reason you think we need to specify the type? MacosColor extends Color, and Dart's type checker is pretty good at determining the type in cases like this.

Code consistency.

GroovinChip:
> > Any particular reason you think we need to specify the type? MacosColor extends Color, and Dart's type checker is pretty good at determining the type in cases like this.
> 
> Code consistency.

I'd then prefer not to declare the type for transparent

example/lib/pages/colors_page.dart
```diff
@@ -0,0 +1,334 @@
...
+              child: ColorBox(
+                color: MacosColors.quaternaryLabelColor,
+              ),
+            ),
+            MacosTooltip(
+              message: 'Quaternary Label Dark',
+              child: ColorBox(
+                color: MacosColors.quaternaryLabelColor.darkColor,
+              ),
+            ),
```
bdlukaa, line 308: There must be a better way to do this

bdlukaa (approved):

example/lib/pages/colors_page.dart
GroovinChip: 🤷‍♂️

PR #145, opened 2021-06-22
Mouse Cursor

### LINKED ISSUE #139: 🐛 Cursors use `SystemMouseCursors.click` instead of `SystemMouseCursors.basic`
**Description**

Clickable macOS components do not use the "click cursor". They have always used the default cursor. This forms a distinction between using the web on macOS and interacting with macOS applications.

So by default, clickable elements in macos_ui should use `SystemMouseCursors.basic` instead of `SystemMouseCursors.click`. I also believe an option should be provided to customize that.

bdlukaa:
Add `mouseCursor` to `MacosThemeData`. By default, on Apple platforms, `MouseCursor.defer` (no cursor) is used. On the other platforms, `SystemMouseCursors.click` is used.

Closes #139 

## Pre-launch Checklist

### DIFF
--- a/lib/src/buttons/back_button.dart
+++ b/lib/src/buttons/back_button.dart
@@ -119,8 +119,10 @@ class _MacosBackButtonState extends State<MacosBackButton>
 
   @override
   Widget build(BuildContext context) {
+    assert(debugCheckHasMacosTheme(context));
+    final MacosThemeData theme = MacosTheme.of(context);
     final bool enabled = widget.enabled;
-    final brightness = MacosTheme.of(context).brightness;
+    final Brightness brightness = theme.brightness;
     final iconColor = brightness == Brightness.dark
         ? CupertinoColors.white
         : CupertinoColors.black;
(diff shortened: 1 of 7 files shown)


GroovinChip:
@wilsonowilson does this look good to you?

wilsonowilson:
Not sure how I feel about having cursor in the theme. Cursors don't have anything to do with theming, and as such, both `ThemeData` and `CupertinoThemeData` have no cursor property. Instead, cursors are provided to the widgets that need them alone.

But I understand the problem the PR is trying to solve... automatically providing a cursor to widgets that need it based on platform. I suggest creating a separate widget that handles this, perhaps an `InheritedWidget` called `CursorConfiguration`. That way we'd be able to call `CursorConfiguration.of(context)` to access things like `clickableCursor?` and `disabledCursor`. LMK what you think.

GroovinChip:
> Not sure how I feel about having cursor in the theme. Cursors don't have anything to do with theming, and as such, both `ThemeData` and `CupertinoThemeData` have no cursor property. Instead, cursors are provided to the widgets that need them alone.
> 
> But I understand the problem the PR is trying to solve... automatically providing a cursor to widgets that need it based on platform. I suggest creating a separate widget that handles this, perhaps an `InheritedWidget` called `CursorConfiguration`. That way we'd be able to call `CursorConfiguration.of(context)` to access things like `clickableCursor?` and `disabledCursor`. LMK what you think.

I agree with you, that seems like a better approach.

PR #157, opened 2021-09-27
generic support wip

### LINKED ISSUE #156: Improvements to `MacosRadioButton`
Currently `MacosRadioButton` does not support generics. It only supports boolean values - checked or unchecked. Compare this to the `Radio` widget found in the material library, however, and you will find that it supports generic types for the value. See the following example:

```dart
ThemeMode _selectedThemeMode = ThemeMode.system;

Radio<ThemeMode>(
  value: _selectedThemeMode,
  groupValue: ThemeMode.system,
  onChanged: (value) {},
),
```

This is what we should aim for with MacosRadioButton, which we currently cannot do.

sachaarbonel:
Closes #156


GroovinChip:
@sachaarbonel any news on this?

GroovinChip:
Tested. It works!

GroovinChip:
LGTM

Issue #173, opened 2022-01-05
💡 Implement transparency effects


cadenkriese:
## Use case

Arguably, the most glaring difference between a flutter app with macos_ui and a native app is the lack of transparency/blurring on the sidebar and alert dialogs.

## Proposal

Implement transparency / frosted glass effects in relevant places.

For the window transparency, https://github.com/alexmercerind/flutter_acrylic can be used.

PR #175, opened 2022-01-16
Sidebar transparency effect

mj-shifu:
This PR adds a transparency effect to the sidebar, as long as system and app brightness match.

Some changes to MainFlutterWindow.swift are necessary, like they were done in the README and the example

## Pre-launch Checklist

### Before PR
before_white
before_black

## After PR
after_light
after_black

### DIFF
--- a/lib/src/layout/window.dart
+++ b/lib/src/layout/window.dart
@@ -1,5 +1,6 @@
 import 'dart:math' as math;
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:macos_ui/src/indicators/scrollbar.dart';
 import 'package:macos_ui/src/layout/content_area.dart';
@@ -95,10 +96,18 @@ class _MacosWindowState extends State<MacosWindow> {
     late Color sidebarBackgroundColor;
     Color dividerColor = theme.dividerColor;
 
-    // Only show blurry, transparent sidebar when platform brightness and app
-    // brightness are the same, otherwise it looks awful
-    if (MediaQuery.of(context).platformBrightness.isDark ==
-        theme.brightness.isDark) {
+    final isMac = !kIsWeb && defaultTargetPlatform == TargetPlatform.macOS;
+
+    // Respect the sidebar color override from parent if one is given
+    if (widget.sidebar?.decoration?.color != null) {
+      sidebarBackgroundColor = widget.sidebar!.decoration!.color!;
+    } else if (isMac &&
+        MediaQuery.of(context).platformBrightness.isDark ==
+            theme.brightness.isDark) {
+      // Only show blurry, transparent sidebar when platform brightness and app
+      // brightness are the same, otherwise it looks awful. Also only make the
+      // sidebar transparent on native Mac, or it will just be flat black or
+      // white.
       sidebarBackgroundColor = Colors.transparent;
     } else if (!theme.brightness.isDark) {
       sidebarBackgroundColor = widget.sidebar?.decoration?.color ??
--- a/lib/src/layout/window.dart
+++ b/lib/src/layout/window.dart
@@ -97,7 +97,8 @@ class _MacosWindowState extends State<MacosWindow> {
 
     // Only show blurry, transparent sidebar when platform brightness and app
     // brightness are the same, otherwise it looks awful
-    if (MediaQuery.of(context).platformBrightness.isDark == theme.brightness.isDark) {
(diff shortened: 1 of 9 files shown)


mj-shifu:
This should close #173.

lib/src/layout/window.dart
```diff
@@ -95,7 +95,12 @@ class _MacosWindowState extends State<MacosWindow> {
...
-    if (!theme.brightness.isDark) {
+    // Only show blurry, transparent sidebar when platform brightness and app
+    // brightness are the same, otherwise it looks awful
+    if (MediaQuery.of(context).platformBrightness.isDark ==
+        theme.brightness.isDark) {
+      sidebarBackgroundColor = Colors.transparent;
+    } else if (!theme.brightness.isDark) {
       sidebarBackgroundColor = widget.sidebar?.decoration?.color ??
           CupertinoColors.systemGrey6.color;
     } else {
```
jmatth: 2 things:

- This doesn't look right on web and Windows (I didn't test Linux). Screenshots below.
- I would expect any color override from the parent to take precedence, but right now it will be overridden by the transparency effect.

I would suggest the following changes to fix both issues:

```suggestion
    final isMac = !kIsWeb && defaultTargetPlatform == TargetPlatform.macOS;
    
    // Respect the sidebar color override from parent if one is given
    if (widget.sidebar?.decoration?.color != null) {
      sidebarBackgroundColor = widget.sidebar!.decoration!.color!;
    } else if (isMac &&
        MediaQuery.of(context).platformBrightness.isDark ==
            theme.brightness.isDark) {
      // Only show blurry, transparent sidebar when platform brightness and app
      // brightness are the same, otherwise it looks awful. Also only make the
      // sidebar transparent on native Mac, or it will just be flat black or
      // white.
      sidebarBackgroundColor = Colors.transparent;
    } else if (!theme.brightness.isDark) {
      sidebarBackgroundColor = widget.sidebar?.decoration?.color ??
          CupertinoColors.systemGrey6.color;
    } else {
      sidebarBackgroundColor = widget.sidebar?.decoration?.color ??
          CupertinoColors.tertiarySystemBackground.darkColor;
}
```

This change is meant to replace line 97 - 109 rather than 97 - 106, but Github won't let me apply the suggestion that far outside the diff 😒.

Web:

image

Windows:

image

jmatth (changes requested):
Caveat on my review: I am not a member of this repo and can't merge. Also thanks for doing this, you just saved me digging through some Apple docs I was not looking forward to 😁.

mj-shifu:
Thank you, I just changed it according to your suggestion.

lib/src/layout/window.dart
```diff
@@ -95,7 +96,20 @@ class _MacosWindowState extends State<MacosWindow> {
...
+      sidebarBackgroundColor = widget.sidebar!.decoration!.color!;
+    } else if (isMac &&
+        MediaQuery.of(context).platformBrightness.isDark ==
+            theme.brightness.isDark) {
+      // Only show blurry, transparent sidebar when platform brightness and app
+      // brightness are the same, otherwise it looks awful. Also only make the
+      // sidebar transparent on native Mac, or it will just be flat black or
+      // white.
+      sidebarBackgroundColor = Colors.transparent;
+    } else if (!theme.brightness.isDark) {
```
jmatth: Sorry to but in again, I just realized the suggestion I sent was an older version of the code. Since we're now checking for `widget.sidebar?.decoration?.color` in the first `if` branch, we don't need to consider it when falling back to the theme. The last two branches can be collapsed into single else clause with a ternary statement:

```suggestion
    } else {
        sidebarBackgroundColor = CupertinoColors.systemGrey6.color;
    }
```

Or if you prefer without the ternary:

```suggestion
    } else if (!theme.brightness.isDark) {
        sidebarBackgroundColor = CupertinoColors.systemGrey6.color;
    } else {
        sidebarBackgroundColor = CupertinoColors.tertiarySystemBackground.darkColor;
    }
```

Both changes are supposed to replace lines 112 - 118 and haven't been run through `dartfmt` because github's suggestion system is very annoying to use.

GroovinChip: @mj-shifu what are your thoughts on @jmatth 's feedback here?

mj-shifu: I agree with @jmatth 's feedback. But because GitHub wouldn't let me merge the changes directly, I created a new commit to fix it.

GroovinChip (approved):
LGTM! Thanks for your contribution, @mj-shifu !

PR #182, opened 2022-03-12
Fix `MacosThemeData` to properly apply user-defined component theme classes.

whiplashoo:
Fixes an issue where user-defined theme properties `pushButtonTheme`, `helpButtonTheme`, and `tooltipTheme` were not passed properly when creating a custom `MacosThemeData`.

Now, a user can customize a theme like this:

```dart
return MacosApp(
  darkTheme: MacosThemeData.dark().copyWith(
    pushButtonTheme: const PushButtonThemeData().copyWith(
      color: MacosColors.appleBrown,
      disabledColor: MacosColors.applePurple,
      secondaryColor: MacosColors.appleMagenta,
    ),
    helpButtonTheme: const HelpButtonThemeData(
      color: MacosColors.appleOrange,
      disabledColor: MacosColors.appleCyan,
    ),
    tooltipTheme: const TooltipThemeData(
      height: 100.0,
    ),
  ),
)
```

and get properly styled `PushButton`, `HelpButton`, and `MacosTooltip` (not shown) widgets:

Screenshot 2022-03-12 at 11 02 05 PM

Also changed a couple of colors in the `example/main.dart`, to remove its Material library dependency. 

## Pre-launch Checklist


whiplashoo:
Got a couple of failing tests, will check again.

whiplashoo:
OK, there is also another issue with the components' `ThemeData` classes, that caused these tests to fail with null-check errors. 

These occur when a user defines the `PushButtonThemeData` for the theme **without explicitly setting all of its constructor properties**. For example, trying to customize only the `color` of `PushButton`s via the theme:

```dart
theme: MacosThemeData.light().copyWith(
  pushButtonTheme: PushButtonThemeData(
    color: MacosColors.appleOrange,
  ),
),
```

causes a CastError (null-check) on [Line 232](https://github.com/GroovinChip/macos_ui/blob/f82570ae2be490dbec28940ec00f95347df669ee/lib/src/buttons/push_button.dart#L232), because `disabledColor` ends up being null (and another one later, because `secondaryColor` also ends up being null).

I believe we want the user to be able to set just `color` and not have to explicitly set the other two properties to their default values. The same happens with `helpButtonTheme`, `tooltipTheme`, `macosIconButtonTheme`, `macosPopupButtonTheme`.

One solution would be to provide the default colors inside the widgets as well, via theme-level properties. This is how Flutter's material theme does it, for example on `FloatingActionButton`s:

```dart
final Color foregroundColor = this.foregroundColor
  ?? floatingActionButtonTheme.foregroundColor
  ?? theme.colorScheme.onSecondary;
final Color backgroundColor = this.backgroundColor
  ?? floatingActionButtonTheme.backgroundColor
  ?? theme.colorScheme.secondary;
final Color focusColor = this.focusColor
  ?? floatingActionButtonTheme.focusColor
  ?? theme.focusColor;
```

This would require to populate `MacosThemeData` with more top-level properties, and then just [initialize the component themes](https://github.com/flutter/flutter/blob/7e9793dee1b85a243edd0e06cb1658e98b077561/packages/flutter/lib/src/material/theme_data.dart#L512) without setting anything.

GroovinChip:
@whiplashoo I'm also seeing these failing tests. The problem is coming from the fact that in our tests we are not passing any values to the theme classes. See for example [line 22](https://github.com/GroovinChip/macos_ui/blob/fix_theme_classes/test/buttons/push_button_test.dart#L22) of `push_button_test.dart`. When pass values to the theme data class, the test passes.

GroovinChip:
All tests should pass now.

whiplashoo:
Indeed, the tests should pass now. However, I am still seeing the null-check error, when initializing a `PushButtonTheme` with just one property, like:

```dart
theme: MacosThemeData.light().copyWith(
  pushButtonTheme: PushButtonThemeData(
    color: MacosColors.appleOrange,
  ),
),
```

Screenshot 2022-03-13 at 6 21 33 PM

Am I missing something?

GroovinChip:
@whiplashoo I'll look into it

GroovinChip:
I'm thinking the problem is we are missing `merge()` methods for `MacosThemeData` and various widget `ThemeData` classes. See [here](https://github.com/GetStream/stream-chat-flutter/blob/develop/packages/stream_chat_flutter/lib/src/stream_chat_theme.dart#L466) and [here](https://github.com/GetStream/stream-chat-flutter/blob/develop/packages/stream_chat_flutter/lib/src/theme/channel_preview_theme.dart#L125) for examples

whiplashoo:
OK, that looks doable. I will try to work on it soon.

GroovinChip:
> OK, that looks doable. I will try to work on it soon.

Thanks!

PR #185, opened 2022-03-17
feat: added padding property to IconButton

edwardaux:
Previously the padding for `IconButton` was hard-coded to `8`. This PR adds the ability to influence it by adding a `padding` property to both the widget or its theme.

Note that I deliberately did NOT update the changelog because I'm not in charge of deciding what goes in what release and/or actually publishing. I figured I'd leave that to @GroovinChip to add whenever he decides to publish a new release.

## Pre-launch Checklist

- [X] I have run `dartfmt` on all changed files 
- [X] I have incremented the package version as appropriate and updated `CHANGELOG.md` with my changes 
- [X] I have added/updated relevant documentation 
- [X] I have run "optimize/organize imports" on all changed files
- [X] I have addressed all analyzer warnings as best I could

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [0.12.3]
+* Added `padding` property to `MacosIconButton` and `MacosIconButtonTheme`.
+
 ## [0.12.2]
 * Fixes `MacosThemeData` to properly apply user-defined `pushButtonTheme`, `helpButtonTheme`, and `tooltipTheme` properties.
 
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -73,7 +73,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.12.2"
+    version: "0.12.3"
   matcher:
     dependency: transitive
     description:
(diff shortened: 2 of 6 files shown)


GroovinChip:
Hi @edwardaux, thanks for your PR. Would you mind providing a use case where this would be useful? Apple's design kits show a padding of 8, which is why the value is hard coded.

edwardaux:
Oops, sorry... I mean to add a bit more info. In my case, I want to have some icon buttons in a fairly tightly constrained part of my screen. On the left is how I want it to look, but if I use the default padding of `8`, then it blows out the spacing of the rows because the icons are high.

Desired padding | Default padding
:---------:|:-----------:
Screen Shot 2022-03-18 at 8 40 33 am | Screen Shot 2022-03-18 at 8 42 26 am

Note that I'm not proposing changing the default padding - it remains at `8`, however, this PR gives the flexibility for users to customise it (in the same way that Apple provides default settings for `NSButton` but provides hooks for me to customise it if I want to look slightly different to the defaults).

GroovinChip:
@edwardaux fair enough, I suppose. Please address the failing package analysis check and I'll review this PR. Additionally, I'd appreciate it if you updated CHANGELOG.md along with the version in pubspec.yaml - there is a reason I include these steps in the PR checklist. I appreciate your deference to me, though.

edwardaux:
👍 Have fixed failing tests and bumped the version number.

GroovinChip:
@edwardaux since #186 has landed I suggest you pull the latest changes - since there is now a `merge()` method you'll need to add your padding update to it. Aside from that, LGTM to me.

PR #186, opened 2022-03-17
Add merge methods to `MacosThemeData` and widget-specific `ThemeData` classes (closes #183)

### LINKED ISSUE #183: 🐛 Crash in PushButton when using custom theme data
**Description**

I've been dabbling with using theming info and created a custom theme data using:
```
MacosThemeData(pushButtonTheme: const PushButtonThemeData(color: Colors.amber))
```

However, this causes a crash when I use a `PushButton` widget on this line: https://github.com/GroovinChip/macos_ui/blob/65b53fa7529d069f5227d3ae521ac7272d74ea34/lib/src/buttons/push_button.dart#L226

Now, I can see that if I provide a value for `secondaryColor` in `PushButtonThemeData` then this problem will go away. However, the fact that this field is defined as an optional in the theme data suggests that this shouldn't be the case. 

Not really sure what the "solution" is. Maybe it is to make that field mandatory, maybe just some documentation, maybe an assertion if it is null, maybe defaulting to the [default theme value](https://github.com/GroovinChip/macos_ui/blob/65b53fa7529d069f5227d3ae521ac7272d74ea34/lib/src/theme/macos_theme.dart#L217).

Anyway, just thought I'd mention it.

whiplashoo:
Adds missing `merge()` methods to `MacosThemeData` and widget-specific `ThemeData` classes, making it possible to use them properly with any number of user-provided custom properties. This issue was previously discussed in #182 and #183, where if the user didn't explicitly set all properties for a `ThemeData` the app crashed with a null-check CastError.

Now, customizing a couple of theme properties is possible. Example:

```dart
darkTheme: MacosThemeData.dark().copyWith(
  canvasColor: MacosColors.selectedTextBackgroundColor,
  pushButtonTheme: const PushButtonThemeData(
    color: MacosColors.applePurple,
  ),
  tooltipTheme: const TooltipThemeData(
    height: 100.0,
  ),
  macosIconButtonTheme: const MacosIconButtonThemeData(
    backgroundColor: MacosColors.appleBlue,
  ),
  macosPopupButtonTheme: const MacosPopupButtonThemeData(
    highlightColor: MacosColors.appleGreen,
  )),
),
```

Also fixed some minor errors/typos in the widget `ThemeData` classes.

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [0.12.2+1]
+* Adds missing `merge` methods to `MacosThemeData` and widget `ThemeData` classes, making it possible to use them properly with any number of user-provided custom properties.
+
 ## [0.12.2]
 * Fixes `MacosThemeData` to properly apply user-defined `pushButtonTheme`, `helpButtonTheme`, and `tooltipTheme` properties.
 
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -73,7 +73,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.12.2"
+    version: "0.12.2+1"
   matcher:
     dependency: transitive
     description:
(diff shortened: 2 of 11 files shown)


GroovinChip:
LGTM

PR #193, opened 2022-04-19
MacosPulldownButton (closes #86)

### LINKED ISSUE #86: PulldownButton
Screen Shot 2021-05-13 at 9 32 03 PM
Screen Shot 2021-05-13 at 9 32 12 PM

whiplashoo:
## Adds a new `MacosPulldownButton` widget, trying to replicate the [Pull-Down button](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/pull-down-buttons/) component from Apple's Guidelines.

**Final result:**

| Dark Theme                                 | Light Theme                                |
| ------------------------------------------ | ------------------------------------------ |
|  |  |
|  |  |
|  |  |

**Screencast:** 

https://user-images.githubusercontent.com/9117427/163968481-24d3ec1a-7f7c-4ea6-8db5-d4dbd327b27b.mov

It features:

- The ability to use it either with text or an icon as the button's title. These are both categorized as Pulldown buttons by Apple, so it made sense to provide this option here as well. Example of a Pulldown button with icon can be found in the [Figma kit](https://www.figma.com/file/M6K5L3GK0WJh6pnsASyVeE/macOS-Big-Sur-UI-Kit?node-id=26%3A5).
- macOS-themed styling, trying to match as closely as possible the native component. Extra-care was taken to use the correct color values for the different states of the button (enabled, hovered, pressed, and disabled).
- A subtle "frosted-glass" effect for the background of the Pulldown menu, using a `BackdropFilter` with blur.
- The ability to also include dividers (`MacosPulldownMenuDivider`) to separate the menu items (`MacosPulldownMenuItem`).
- Setting items to an empty list ([]) will disable the button (greyed out).

User can also navigate the open menu with up/down arrow keys and select an item with the Return key.

An area where we cannot replicate behavior completely, is the menu's placement, which in the native component may also be drawn beyond the window's limits.

Since Flutter [cannot draw outside the active window](https://github.com/flutter/flutter/issues/30701), we cannot have the menu exceed the available window height or width. So, the menu will always appear right below the button, except if there is no available vertical space, when it will appear above the button.

## Pre-launch Checklist

- [✅] I have run `dartfmt` on all changed files 
- [✅] I have incremented the package version as appropriate and updated `CHANGELOG.md` with my changes 
- [✅] I have added/updated relevant documentation 
- [✅] I have run "optimize/organize imports" on all changed files
- [✅] I have addressed all analyzer warnings as best I could


whiplashoo:
Two more notes:

- Customizing the `MacosPulldownButtonTheme` is also possible. So, setting it to something like this:

```dart
return MacosApp(
    title: 'macos_ui example',
    theme: MacosThemeData.light().copyWith(
    macosPulldownButtonTheme: const MacosPulldownButtonThemeData(
        highlightColor: MacosColors.appleOrange,
        backgroundColor: MacosColors.appleGreen,
        pulldownColor: MacosColors.systemTealColor,
    ),
    ),
```

will build this button:
Screenshot 2022-04-19 at 10 12 57 AM

- While working on this, I found some features and details that can be added to the `MacosPopupButton` as well, like the frosted glass effect, or some unnecessary properties. I will be opening a PR with improvements to that soon.

GroovinChip (approved):
LGTM

PR #240, opened 2022-05-09
Fix toolbar `actions` RangeError (closes #239)

### LINKED ISSUE #239: 🐛  Changing the number of toolbar items (`actions`) programmatically throws a `RangeError`
**Description**

While integrating the new toolbar into my app, I came across a `RangeError` when changing programmatically the number of toolbar items (`actions`). Screencast:

https://user-images.githubusercontent.com/9117427/167438342-95ea501b-1516-4a15-b15b-d9a7153c7c06.mov

**Steps To Reproduce**

1. Create a `Toolbar` with any number of `actions`.
2. Resize the window so that the toolbar is overflowed.
3. Then try changing the number of `actions` provided to the toolbar. In my app, this is done when changing views (from home screen to settings), because the settings screen has only one toolbar item (instead of 7). It can also be tested by commenting out a Toolbar item.
4. A `RangeError` is thrown from here: https://github.com/GroovinChip/macos_ui/blob/3738bd4855ac3937c3d3691291ce3a00cf9e1a7b/lib/src/layout/toolbar/toolbar.dart#L179  This happens because while the toolbar actions have changed, the overflowed actions list is not updated as well, but is stuck on the previous state.

**Expected behavior**
 
Fixing this will require observing changes to the toolbar's `actions` property and forcing a recalculation of the oveflowed actions list.

whiplashoo:
Fixes the error thrown when modifying the `actions` list of a toolbar after initialization.

Adds a `didUpdateWidget` method to the `Toolbar`, in order to track modifications to the widget's `actions` property. If it has changed for any reason, it forces a recalculation of the overflowed widgets.

Also simplifies how we track which actions have overflowed. Instead of checking their indexes, we now just count how many of them (`overflowedActionsCount`) should be hidden, and slice the end of the `actions` array accordingly.

Closes #239.

## Pre-launch Checklist

- [👍 ] I have run `dartfmt` on all changed files 
- [ 👍 ] I have incremented the package version as appropriate and updated `CHANGELOG.md` with my changes 
- [] I have added/updated relevant documentation 
- [ 👍 ] I have run "optimize/organize imports" on all changed files
- [ 👍 ] I have addressed all analyzer warnings as best I could

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [1.2.1]
+* Fixes issue with error thrown when toolbar actions are modified programmatically [#239](https://github.com/GroovinChip/macos_ui/issues/239) 
+
 ## [1.2.0]
 * Improved styling for `MacosTooltip`:
   * Better color and shadows.
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -87,7 +87,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "1.2.0"
+    version: "1.2.1"
   matcher:
     dependency: transitive
     description:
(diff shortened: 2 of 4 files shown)


GroovinChip (changes requested):
LGTM, just run a pub get so pubspec.lock updates with the new version number please.

PR #260, opened 2022-05-29
remove setState call in the ResizablePane didUpdateWidget method

tonyborchert:
Removing as documentation states, that after didUpdateWidget is called, setState is also called.

The call is therefore unnecessary and should be removed.


GroovinChip:
@bit-burger Thanks for the PR. What "documentation" are you referring to?

GroovinChip:
Closing due to inactivity.

PR #261, opened 2022-05-29
disabled background color force

nbonamy:
Needed to force disabled background color

### DIFF
--- a/lib/src/fields/text_field.dart
+++ b/lib/src/fields/text_field.dart
@@ -236,6 +236,7 @@ class MacosTextField extends StatefulWidget {
     this.focusNode,
     this.decoration = kDefaultRoundedBorderDecoration,
     this.focusedDecoration = kDefaultFocusedBorderDecoration,
+    this.disabledColor,
     this.padding = const EdgeInsets.all(4.0),
     this.placeholder,
     this.placeholderStyle = const TextStyle(
@@ -370,6 +371,7 @@ class MacosTextField extends StatefulWidget {
     this.focusNode,
     this.decoration,
     this.focusedDecoration,
+    this.disabledColor,
     this.padding = const EdgeInsets.fromLTRB(2.0, 4.0, 2.0, 4.0),
     this.placeholder,
     this.placeholderStyle = _kDefaultPlaceholderStyle,
@@ -479,6 +481,11 @@ class MacosTextField extends StatefulWidget {
   /// Defaults to having a rounded rectangle blue border and can be null to have
   /// no box decoration.
   final BoxDecoration? focusedDecoration;
+
+  /// Controls the background color of disabled text field.
+  ///
+  /// Defaults is to use system disabled color.
+  final Color? disabledColor;
 
   /// Padding around the text entry area between the [prefix] and [suffix]
   /// or the clear button when [clearButtonMode] is not never.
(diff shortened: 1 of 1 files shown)


GroovinChip:
@nbonamy Thanks for the PR. Can you please provide some more information on your changes? What did you need this for? Can you provide a screenshot or gif that demonstrates your change?

nbonamy:
Hello,

I am writing am image browser and viewer (https://github.com/nbonamy/foto) with functionalities similar to Finder. Hence I am using MacosTextField for the names of the files that I set to disabled. When user selects "Rename", I enable the field.

Still struggling with focus though...

SCR-20220530-kf7
SCR-20220530-kfg

Thanks

GroovinChip:
Couldn't you achieve this effect simply by setting `enabled: false` and setting a custom `decoration` until the user clicks "rename"?

nbonamy:
No because line 1311, effectiveDecoration is built with color being overridden with the calculated disabledColor if widget is disabled. I went for the simplest fix as I did not want to mess too much with the codebase. There maybe a more elegant solution 😀

GroovinChip:
@whiplashoo what are your thoughts on this?

whiplashoo:
OK, I think that enabling changing the disabled background color of a `MacosTextField` sounds reasonable.

However, @nbonamy in your specific use case, I think it would be best to either:

- Use two different widgets for the name label, one when user is renaming (`MacosTextField`) and one when not (simple `Text`). When user double clicks on the text, the text field is shown instead. Example code:

```dart
bool isRenaming = false;
TextEditingController renameController = TextEditingController()
  ..text = "Test name";

SizedBox(
  width: 300.0,
  child: isRenaming
      ? MacosTextField(
          controller: renameController,
        )
      : GestureDetector(
          onDoubleTap: () => setState(() => isRenaming = true),
          child: Text(renameController.text)),
  ),
```

It would be more natural I think; having a disabled text field for the label seems a bit hacky. It's probably what MacOS Finder does under the hood, and you can also have more control over the labels when inactive. 

- Another solution could be playing with the `readonly` attribute of the `MacosTextField` instead of `enabled`. Setting `readonly` to true only removes the gesture detection and you can set a transparent decoration too:

```dart
SizedBox(
  width: 300.0,
  child: GestureDetector(
    onDoubleTap: () => setState(() => isRenaming = true),
    child: MacosTextField(
      readOnly: isRenaming ? false : true,
      controller: renameController,
      decoration: BoxDecoration(
        border: Border.all(color: Colors.transparent),
      ),
    ),
  ),
),
```

Please have a look at these suggestion in case they help with your project!

nbonamy:
Thanks! Will look into doing differently.

GroovinChip:
Closing until further notice

nbonamy:
Thought that your comment "OK, I think that enabling changing the disabled background color of a MacosTextField sounds reasonable." meant you were going to accept the PR 😀

GroovinChip:
Later discussion indicates that there are other methods of achieving this, which you are going to look into. If nothing else pans out, update me here and I'll re-open the PR.

nbonamy:
Hi. Tried both solutions.

1) Going back and forth from Text to MacosTextField seems to pose some problem. Specifically, getting "setState() called after dispose(): _MacosTextFieldState" from _handleFocusChanged. Maybe something is wrong with my implementation but I kept it as simple as possible

2) readonly shows a caret cursor which is not desirable

Regards
Nicolas

PR #206, opened 2022-04-24
MacosTabView & MacosTab

Andrewngabriel:
Add Tabview

Closes #70

Demo:

https://user-images.githubusercontent.com/11186948/166086219-3fc9c6c1-8802-46a7-b4ba-93b47814270c.mov

## Pre-launch Checklist

- [x] I have run `dartfmt` on all changed files 
- [ ] I have incremented the package version as appropriate and updated `CHANGELOG.md` with my changes 
- [x] I have added/updated relevant documentation 
- [x] I have run "optimize/organize imports" on all changed files
- [x] I have addressed all analyzer warnings as best I could

PR #273, opened 2022-07-04
Version 1.6.0 - `MacosTabView` & `MacosSegmentedControl`

### LINKED ISSUE #70: `MacosTabView` & `MacosTab`
https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/tab-views/

A tab view presents multiple mutually exclusive panes of content in the same area. A tab view includes a tabbed control and a content area. Each segment of a tabbed control is known as a tab, and clicking a tab displays its corresponding pane in the content area. Although the amount of content can vary from pane to pane, switching tabs doesn’t change the overall size of the tab view or its parent window.

TabView Preview

| Vertical Tabs | Horizontal Tabs |
| ------------- | ----------------- |
|  |  |
|  |  |

GroovinChip:
This PR introduces two new widgets: `MacosTabView` and `MacosSegmentedControl`. Relevant tests have been added.

It also contains improvements to overall dartdocs.

Closes #70.

See [this comment](https://github.com/GroovinChip/macos_ui/pull/206#pullrequestreview-958523992) from #206 for background info.

I'm not super pleased with the API here, but it's much closer to what it should be. The thing that I'm unhappy about is the `active` property for `MacosTab`; I feel that it should be managing its own state rather than depending on it via class properties. This will allow multiple states for tabs, and we know that the native control has more states than simply "active" or "inactive". I tried using `MaterialStateProperty` but got nowhere. I also do not want to use a `color` parameter because the colors of that tab and the tab's states should only be set internally. Putting the `GestureDetector` that handles the tab click into the tab widget itself is also problematic because it requires functions and/or the tab controller to be passed into the tab, and that's not correct. 

The above issue can either be cleaned up in this PR, or in a separate PR.

## Pre-launch Checklist

### DIFF
--- a/example/lib/pages/tabview_page.dart
+++ b/example/lib/pages/tabview_page.dart
@@ -1,4 +1,4 @@
-import 'package:flutter/material.dart';
+import 'package:flutter/cupertino.dart';
 import 'package:macos_ui/macos_ui.dart';
 
 class TabViewPage extends StatefulWidget {
--- a/example/lib/pages/tabview_page.dart
+++ b/example/lib/pages/tabview_page.dart
@@ -1,0 +1,62 @@
+import 'package:flutter/cupertino.dart';
+import 'package:macos_ui/macos_ui.dart';
+
+class TabViewPage extends StatefulWidget {
+  const TabViewPage({super.key});
+
+  @override
+  State<TabViewPage> createState() => _TabViewPageState();
+}
+
+class _TabViewPageState extends State<TabViewPage> {
+  final _controller = MacosTabController(
+    initialIndex: 0,
+    length: 3,
+  );
+
+  @override
+  Widget build(BuildContext context) {
+    return MacosScaffold(
+      toolBar: const ToolBar(
+        title: Text('TabView'),
+      ),
+      children: [
+        ContentArea(
+          builder: (context, scrollController) {
+            return Padding(
+              padding: const EdgeInsets.all(24.0),
+              child: MacosTabView(
+                controller: _controller,
+                tabs: [
+                  MacosTab(
+                    label: 'Tab 1',
+                    active: _controller.index == 0,
+                  ),
+                  MacosTab(
+                    label: 'Tab 2',
+                    active: _controller.index == 1,
+                  ),
+                  MacosTab(
+                    label: 'Tab 3',
+                    active: _controller.index == 2,
+                  ),
+                ],
+                children: const [
+                  Center(
+                    child: Text('Tab 1'),
+                  ),
+                  Center(
+                    child: Text('Tab 2'),
+                  ),
+                  Center(
+                    child: Text('Tab 3'),
+                  ),
+                ],
+              ),
+            );
+          },
+        ),
+      ],
+    );
+  }
+}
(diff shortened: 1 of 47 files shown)


GroovinChip:
@whiplashoo any chance of a review?

whiplashoo:
Yes, I am on a trip now, so I'll get to it by the end of the week, if that's okay

GroovinChip:
> Yes, I am on a trip now, so I'll get to it by the end of the week, if that's okay
> 
> 

Sure, no problem

example/lib/pages/tabview_page.dart
```diff
@@ -1,4 +1,4 @@
-import 'package:flutter/cupertino.dart';
```
GroovinChip: Why the change to material from cupertino?

whiplashoo:
Regarding the `active` property issue:

I also think it shouldn't need to be set by the user. So, I initially thought to make it non-required, and handle it internally via the tab controller. However, I found that this is already done by the `MacosSegmentedControl`'s build method [here](https://github.com/GroovinChip/macos_ui/blob/e1056ec5d072f1bd426d154955e171e76d82385b/lib/src/buttons/segmented_control.dart#L76). Making the `active` property optional works, so that the user won't need to set it explicitly.

Regarding the default colors:

The current tab colors (tab buttons and background) have a purplish hue. Shouldn't it be the more neutral colors, as seen in the default macOS app (build via XCode) here:

Screenshot 2022-07-09 at 10 10 26 PM

I could give this a shot, if you agree.

example/lib/pages/tabview_page.dart
whiplashoo: My bad, it auto imported material when I was tinkering with the Tabs widgets. Will fix.

GroovinChip:
@whiplashoo yes, I agree that the colors are not correct. Feel free to change them.

whiplashoo:
Screenshot 2022-07-09 at 10 54 32 PM
Screenshot 2022-07-09 at 10 54 24 PM

Fixed colors and borders. Will see the rest of changes and then merge.

GroovinChip:
Let me know your thoughts on how the state of tabs should be managed, too. Holding down a tab that isn't selected should highlight it, and so should holding down the currently selected tab.

whiplashoo (approved):
LGTM! Nice job with the documentation updates as well.

whiplashoo:
>  Let me know your thoughts on how the state of tabs should be managed, too. Holding down a tab that isn't selected should highlight it, and so should holding down the currently selected tab.

Yes, I just noticed that this behavior exists too. Should we see it in a future PR? I believe most of the functionality and styling of this widget works fine via this one.

GroovinChip:
> > Let me know your thoughts on how the state of tabs should be managed, too. Holding down a tab that isn't selected should highlight it, and so should holding down the currently selected tab.
> 
> Yes, I just noticed that this behavior exists too. Should we see it in a future PR? I believe most of the functionality and styling of this widget works fine via this one.

I really don't know what the right way to handle this is, so yeah let's do it in a future PR.

PR #298, opened 2022-08-31
fix: MacosWindow's background color

wurikiji:
The current version displays the background color only behind the content area. The background should be laid everywhere to show the `Sidebar`'s color property. 

I changed `AnimatedPositioned` to `SizedBox` and `ColoredBox` because animation is unnecessary. 

## Pre-launch Checklist

## AS-IS

image

## TO-BE

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,375 +1,463 @@
+## [1.7.5]
+
+- Fixed an unappropriate `MacosWindow`'s background color
+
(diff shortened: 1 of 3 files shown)


wurikiji:
This is the wrong approach. I found that this issue is not related to this. 
This problem only appears when the system's brightness theme and the flutter's brightness theme mismatch. (using the toggle brightness command with the flutter tool)

PR #363, opened 2023-02-18
fix(docs): Remove "+" character from example

cbenhagen:
Most likely a copy paste leftover. Hopefully no changelog entry needed.


GroovinChip:
@cbenhagen this was actually purposeful. I intended it to demonstrate that line as being highlighted as an addition, but it clearly didn't work 😅

GroovinChip:
Kinda like [this](https://gist.github.com/salmedina/ad8bea4f46de97ea132f71b0bca73663#file-markdowndiffexample-md)

cbenhagen:
Oh ok. Feel free to close this then. I would argue to take it out so it doesn't create a problem when copy pasting.

GroovinChip:
Well, if you know how to make it look the way it should that would be great. Otherwise, yeah, I guess it should be removed.

schilken:
I also just copy&pasted the Swiftcode and got this compile error:
`/macos/Runner/MainFlutterWindow.swift:32:1: error: binary operator '+' cannot be applied to operands of type 'NSRect' (aka 'CGRect') and 'NSColor?'`

I think many Flutter developers will just copy&paste the code.

Maybe wrap the '+' in a comment like /* added */ or better: just remove it. +1

cbenhagen:
@GroovinChip why would you prefer to format this as a diff? Shouldn't the code snippet reflect the currently needed code for Flutter stable? History is always available if someone needs to see the changes that were made.

GroovinChip:
> @GroovinChip why would you prefer to format this as a diff? Shouldn't the code snippet reflect the currently needed code for Flutter stable? History is always available if someone needs to see the changes that were made.

Yeah, that's fair.

Thank you all for setting me straight on this!

PR #377, opened 2023-03-02
Migrate to macos_window_utils

### LINKED ISSUE #315: No wallpaper tint support
I wish there were support for Monterey background colors:

image

image

Adrian-Samoticha:
This PR migrates to [macos_window_utils](https://pub.dev/packages/macos_window_utils). The benefits of the migration are documented in [this discussion](https://github.com/macosui/macos_ui/discussions/322).

**Short recap:**

+ Fixes #315.
+ Improves animation smoothness while animating windows.
+ Fixes some visual artifacts.
+ Allows the sidebar to remain transparent even when the brightness setting mismatches the OS setting.
+ Changes the default `canvasColor` ~~to mimic [`NSColor.windowBackgroundColor`](https://developer.apple.com/documentation/appkit/nscolor/1528630-windowbackgroundcolor)~~ to color values captured using the Digital Color Meter.

**Drawbacks:**

The implementation is not perfect because Flutter [`ImageFilter.blur`](https://api.flutter.dev/flutter/dart-ui/ImageFilter/ImageFilter.blur.html) doesn't support transparency and it is currently not possible to implement custom image filters (see:  https://github.com/flutter/flutter/issues/16296). This makes it impossible to place a blurry backdrop filter on top of a wallpaper-tinted area since those are implemented by placing an [`NSVisualEffectView`](https://developer.apple.com/documentation/appkit/nsvisualeffectview) underneath a transparent Flutter widget.

The workaround is to disable wallpaper tinting whenever a blurry backdrop filter is present in the widget tree. The `MacosWindow` widget now provides a `disableWallpaperTinting` property if this behavior is undesired. Additionally, the `SliverToolBar` and `ToolBar` widgets provide an `allowWallpaperTintingOverrides` property which can be used to disable their ability to deactivate wallpaper tinting when they are currently not visible on the screen (e.g. when they are hidden by an `IndexedStack`).

Furthermore, I would like to request @GroovinChip to test the changes and make sure they are working as intended. In particular, I have not tested the package's web support, though I assume enabling the `disableWallpaperTinting` should suffice in making the package work.

Additionally, I would like to stress that this is a breaking change. It is my understanding that semantic versioning demands the major version to be incremented in this case, hence the new version is `2.0.0`. This should also prevent the Dart package manager from automatically updating to this version, thus breaking people's builds. However, I have noticed that breaking changes have previously been introduced by only incrementing the minor version. If this is preferred, you are free to change the version to `1.13.0` instead.

Lastly, I would like to ask if it is OK that macos_ui now requires macos_window_utils to be set up properly in the `MainFlutterWindow.swift` file to work properly. Previously, macos_ui worked fine without it, only the window's appearance suffered as a result of the setup being omitted. Now it's likely that macos_window_utils will print warnings to the console if it detects that it hasn't been set up properly. If this is undesired, I'd be happy to make changes to both macos_ui and macos_window_utils to be able to detect whether the user made the necessary changes to the `MainFlutterWindow.swift` file and disable certain features if that is the case.

## Pre-launch Checklist


GroovinChip:
Hi @Adrian-Samoticha!! Wow, thanks so much for all your hard work on this! I'll address a few of your points, and ask some questions of my own, before we begin the review process.

### Package Version
You are correct that sometimes I have introduced breaking changes as minor version increments. This is because I have felt in those instances that while the changes were breaking, they were small enough that a major version change was not warranted. In this case, however, I feel that a major version increment is appropriate.

### Changes to `MainFlutterWindow.swift`
Yes, I believe that updating `MainFlutterWindow.swift` to explicitly use `macos_window_utils` is appropriate. Please feel free to do that.

### Color Correctness
I have been auditing `macos_ui` for accuracy, and a major part of that effort is examining our use of colors. I have found that many of them are incorrect. After a lot of research and banging my head against the wall, I have discovered that the macOS translucency effect is the culprit here, so in order to ensure color correctness I have turned off the transparency effect through the system Accessibility settings. I was then able to recreate native colors accurately. 

However, the next phase of this effort provides a challenge, which I am wondering if `macos_window_utils` can help out with. The challenge is that the colors I have successfully recreated have fully opaque alpha values, and when the transparency effect is on, the colors become incorrect when compared to their counterparts in a native app. 

This leaves me with some questions:
1) As far as _I_ know, there is no way to retrieve the specific value that macOS uses to apply transparency in AppKit. Do you know of a way to do this? If so, can `macos_window_utils` retrieve this value so that `macos_ui` we can apply it to our colors?
2) Another solution, and this might be something that can run parallel with the solution to the first question, would be for `macos_window_utils` to provide something like an `isTransparencyOn` getter which would check if A) system-wide transparency is on, and B) has the Flutter app enabled transparency through `MainFlutterWindow.swift`. `macos_ui` could check this value when `MacosApp` starts up, or if doing that is too much of a performance hit, we could inform developers to run the method that checks for this value in their `main()` and provide it to `MacosApp`, although at that point the performance hit would likely be the same, since there would be an `async` call to native in either case.

Color correctness is really important, and my research has shown that the colors of many `AppKit` components are affected when system transparency is on. If we can resolve this with `macos_window_utils`, then I'd like to do so as part of the integration process, rather than waiting until after.

### An Updated `MacosSwitch` widget
As part of the audit I mentioned previously, I decided to finally stop wrapping `CupertinoSwitch` and rewrite `MacosSwitch` from the ground up, using `CupertinoSwitch`'s code as a guide. I have successfully done this, but its parity with the switch component from AppKit is based in the color accuracy work I mentioned above. If we are able to solve that problem with `macos_window_utils`, then I would like to include this widget in the `2.0.0` release. Is that okay with you?

**EDIT:** I have not pushed any of the code related to color correctness and `MacosSwitch`. If you'd like me to so you can take a look, please let me know and I will do so!

Adrian-Samoticha:
> 1. As far as _I_ know, there is no way to retrieve the specific value that macOS uses to apply transparency in AppKit. Do you know of a way to do this? If so, can `macos_window_utils` retrieve this value so that `macos_ui` we can apply it to our colors?

It cannot, but it might be able to in the future. This open issue might be of interest: https://github.com/Adrian-Samoticha/macos_window_utils.dart/issues/15. Assuming the colors you need are listed in [UI Element Colors](https://developer.apple.com/documentation/appkit/nscolor/ui_element_colors) it will very much be possible to just retrieve the colors at runtime, and this even includes colors that change depending on the user's system preferences.

> 2. Another solution, and this might be something that can run parallel with the solution to the first question, would be for `macos_window_utils` to provide something like an `isTransparencyOn` getter which would check if A) system-wide transparency is on, and B) has the Flutter app enabled transparency through `MainFlutterWindow.swift`. `macos_ui` could check this value when `MacosApp` starts up, or if doing that is too much of a performance hit, we could inform developers to run the method that checks for this value in their `main()` and provide it to `MacosApp`, although at that point the performance hit would likely be the same, since there would be an `async` call to native in either case.

You can check if the user has enabled the “Reduce transparency” setting using `NSWorkspace.shared.accessibilityDisplayShouldReduceTransparency` in AppKit, so that could absolutely be implemented into macos_window_utils. The performance of that shouldn't be an issue. Platform channels are fast enough that I use them to animate the opening and closing animation of the sidebar through Dart, so checking it once shouldn't make a difference at all.

That said, I assume this won't be necessary for the color accuracy issue, since getting the colors directly from [UI Element Colors](https://developer.apple.com/documentation/appkit/nscolor/ui_element_colors) *should* give us perfectly accurate colors, however, we absolutely should disable all Flutter-rendered transparency effects (such as those in the context menu or the toolbar) when that setting is enabled, so adding that into macos_window_utils definitely wouldn't hurt. In fact, the same goes for all accessibility features. That said, I'd suggest we worry about that after the 2.0.0 release and not get carried away by feature creep.

> ### An Updated `MacosSwitch` widget
> As part of the audit I mentioned previously, I decided to finally stop wrapping `CupertinoSwitch` and rewrite `MacosSwitch` from the ground up, using `CupertinoSwitch`'s code as a guide. I have successfully done this, but its parity with the switch component from AppKit is based in the color accuracy work I mentioned above. If we are able to solve that problem with `macos_window_utils`, then I would like to include this widget in the `2.0.0` release. Is that okay with you?
> 
> **EDIT:** I have not pushed any of the code related to color correctness and `MacosSwitch`. If you'd like me to so you can take a look, please let me know and I will do so!

Yeah, that's fine. I was wondering why the switch was so much larger than the one I'm used to on macOS, so I guess that explains it. :P

GroovinChip:
> It cannot, but it might be able to in the future. This open issue might be of interest: https://github.com/Adrian-Samoticha/macos_window_utils.dart/issues/15. Assuming the colors you need are listed in [UI Element Colors](https://developer.apple.com/documentation/appkit/nscolor/ui_element_colors) it will very much be possible to just retrieve the colors at runtime, and this even includes colors that change depending on the user's system preferences.

Yes, the colors are in that list. However, they get affected by the transparency and I am not certain that this happens upon color retrieval or sometime during the creation/rendering of the component. Also, retrieving the colors through native requires an async call, which is incompatible with how `MacosColors` provides static const colors (I've tried making it work, and it doesn't), and I'd prefer to keep that API. I'm not thrilled with the idea of retrieving colors every time an app starts up. The only time I think that would be necessary is for the system accent color.

> You can check if the user has enabled the “Reduce transparency” setting using NSWorkspace.shared.accessibilityDisplayShouldReduceTransparency in AppKit, so that could absolutely be implemented into macos_window_utils. The performance of that shouldn't be an issue. Platform channels are fast enough that I use them to animate the opening and closing animation of the sidebar through Dart, so checking it once shouldn't make a difference at all.

Sounds good! I think we should do that.

> getting the colors directly from [UI Element Colors](https://developer.apple.com/documentation/appkit/nscolor/ui_element_colors) should give us perfectly accurate colors

Again, would getting the colors directly give us the colors with correct alpha values when transparency is on? Hard to know. Maybe we can run an experiment and see 🤔 

> however, we absolutely should disable all Flutter-rendered transparency effects (such as those in the context menu or the toolbar) when that setting is enabled, so adding that into macos_window_utils definitely wouldn't hurt. In fact, the same goes for all accessibility features.

Agreed

> That said, I'd suggest we worry about that after the 2.0.0 release and not get carried away by feature creep.

I'd like to support both transparency and non-transparency in the same release, it makes me uncomfortable to do one and not the other. Unless I'm misunderstanding what you're saying here?

Adrian-Samoticha:
I wouldn't worry too much about the async call. I assume if we were to retrieve the colors during startup and store them in a variable we'd be fine. At worst we'd need to trigger a rebuild once the colors have been retrieved. Hard-coding the colors is something that Apple's [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/foundations/color#dynamic-system-colors) actually advise against, though admittedly we do operate under different conditions being a Flutter app rather than a native one.

**EDIT:** I just realized that since we do offer web support, we need to at least provide hard-coded default values, so maybe you're right.

That said, can you show me a screenshot depicting the color difference depending on the “Reduce transparency” setting you observed? I'd just like to make sure we're on the same page here. For the record, I noticed that the background color changes when the setting is activated, though interestingly `NSColor.windowBackgroundColor` remains the same. I'm not sure if that means that colors provided by `NSColor` don't respect that setting or if macOS simply uses some other color value if the setting is enabled.

> I'd like to support both transparency and non-transparency in the same release, it makes me uncomfortable to do one and not the other. Unless I'm misunderstanding what you're saying here?

What I'm saying is that we should detect the accessibility settings the user has set and react accordingly. In the case of “Reduce transparency,” we should disable the backdrop filter of the overlay filter and the sliver toolbar if we find it to be enabled, which is a feature we currently don't support. I'd argue that we should limit our scope for the 2.0.0 release, though, since respecting accessibility settings may be a little too involved for the time being. The color accuracy issue is still within our scope, though.

GroovinChip:
Here's some screenshots to demonstrate the color difference issue:
| SwiftUI| Flutter | Transparency on |
|-----------|-----------|--------------------|
| image | image | ❌ |
| image | image | ✅ |

Note how in the native app, the track color of the switch is different with transparency on, while in the Flutter app the track color is the same regardless of transparency being on or off. I believe that a transparency effect is being applied to the native switch at some point, but I'm not sure if it's part of the code for that component, or if its the dynamic color doing it, or simply the system itself. (Don't mind the difference in color value between native and Flutter when transparency is off - I thought I had nailed that down but apparently not.)

Sure, we can account for the accessibility setting in a later release.

Adrian-Samoticha:
I believe what you are observing is just wallpaper tinting being disabled when “Reduce transparency” is on, since the “Reduce transparency” setting automatically disables wallpaper tinting. I have disabled wallpaper tinting on my system and found that I could no longer observe that effect. That said, I couldn’t find the switch’s background color in the UI Element Colors list, so I tried to find its (transparent) color myself. It seems that in dark mode its color is `#13ffffff` and in light mode it is `#13ffffff`. In other words, it’s just either fully white or black with an alpha of 19.

If you wish you can push your switch’s code so I can check it out myself and see if the colors I have determined work well with wallpaper tinting enabled as well.

GroovinChip:
I've observed this effect in other components as well, not just the switch. I just used the switch as an example. It happens to buttons as well. Why would wallpaper tinting affect components too? It doesn't make sense.

Adrian-Samoticha:
I assume it’s because those components are slightly transparent. Can you disable wallpaper tinting on your system and see if you can still observe the difference?

GroovinChip:
That's exactly what I'm showing above.

Adrian-Samoticha:
Wallpaper tinting is still enabled in your screenshots. I mean that you should disable the “Allow wallpaper tinting in windows” setting in the “Appearance” tab of your System Settings:

image

I assume that toggling the “Reduce transparency” toggle will have no effect on the background color when “Allow wallpaper tinting in windows” is disabled.

GroovinChip:
> Wallpaper tinting is still enabled in your screenshots. I mean that you should disable the “Allow wallpaper tinting in windows” setting in the “Appearance” tab of your System Settings:
> 
> image
> 
> I assume that toggling the “Reduce transparency” toggle will have no effect on the background color when “Allow wallpaper tinting in windows” is disabled.

Oh interesting. I'll check that out and report back. Thanks!

GroovinChip:
@Adrian-Samoticha OK so that helps somewhat. Here's a comparison table of color value changes that comes from enabling and disabling wallpaper tinting, using a native switch in dark mode as an example.

| Wallpaper tinting off | Wallpaper tinting on |
|----------------------|---------------------|
| image | image |

The question now is how do we determine the right amount of alpha value to apply to components that are affected by wallpaper tinting in order to reach the same RGB values as native components? Best case, If there's a single value that all native components use when wallpaper tinting is on, then we can easily apply that value using `withOpacity` to all colors. If not, we'll need to determine the right amount of opacity per-component.

Adrian-Samoticha:
For the switch, I just assumed that the color is either fully black or white with some given opacity, which made it easy to determine the opacity through trial and error. However, other components may have different colors, so that approach isn’t going to get us very far.

Therefore, I tried a more mathematical approach. The problem I needed to solve is that given two different background colors (i.e. one with wallpaper tinting on and one with it off) and the corresponding rendered component colors, what is the original color and opacity of the component?

So, let $b_1, b_2 \in \mathbb R^3$ be our two background colors and $c_1, c_2 \in \mathbb R^3$ the captured component colors when rendered on top of their respective background colors.

We are looking for the component color $\gamma \in \mathbb R^3$ and its opacity $\alpha \in \mathbb R$, such that

$$\begin{aligned}
{b_1}_i + \alpha(\gamma_i - {b_1}_i) &= {c_1}_i \ \wedge \\
{b_2}_i + \alpha(\gamma_i - {b_2}_i) &= {c_2}_i \\
\Leftrightarrow \quad \quad
\alpha(\gamma_i - {b_1}_i) &= {c_1}_i - {b_1}_i \ \wedge \\
\alpha(\gamma_i - {b_2}_i) &= {c_2}_i - {b_2}_i \\
\Leftrightarrow \quad \quad
\alpha &= \frac{{c_1}_i - {b_1}_i}{\gamma_i - {b_1}_i} \ \wedge \\
\gamma_i &= \alpha^{-1} ({c_2}_i - {b_2}_i) + {b_2}_i \\
\end{aligned}$$

for $i \in \{1, 2, 3\}.$

Solving for $\gamma$ / $\alpha$ we get:

$$\begin{aligned}
\Rightarrow \gamma_i &= \frac{\gamma_i - {b_1}_i}{{c_1}_i - {b_1}_i} ({c_2}_i - {b_2}_i) + {b_2}_i \\
&= \frac{(\gamma_i - {b_1}_i)({c_2}_i - {b_2}_i)}{{c_1}_i - {b_1}_i} + {b_2}_i \\
\Leftrightarrow \gamma_i - {b_2}_i &= \frac{(\gamma_i - {b_1}_i)({c_2}_i - {b_2}_i)}{{c_1}_i - {b_1}_i} \\
\Leftrightarrow (\gamma_i - {b_2}_i)({c_1}_i - {b_1}_i) &= (\gamma_i - {b_1}_i)({c_2}_i - {b_2}_i) \\
\Leftrightarrow \gamma_i ({c_1}_i - {b_1}_i) - {b_2}_i ({c_1}_i - {b_1}_i) &= \gamma_i ({c_2}_i - {b_2}_i) - {b_1}_i ({c_2}_i - {b_2}_i) \\
\Leftrightarrow \gamma_i ({c_1}_i - {b_1}_i) - \gamma_i ({c_2}_i - {b_2}_i) &= {b_2}_i ({c_1}_i - {b_1}_i) - {b_1}_i ({c_2}_i - {b_2}_i) \\
\Leftrightarrow \gamma_i (({c_1}_i - {b_1}_i) - ({c_2}_i - {b_2}_i)) &= {b_2}_i{c_1}_i - {b_2}_i{b_1}_i - {b_1}_i{c_2}_i + {b_1}_i{b_2}_i \\
\Leftrightarrow \gamma_i ({c_1}_i + {b_2}_i - {b_1}_i - {c_2}_i) &= {b_2}_i{c_1}_i - {b_1}_i{c_2}_i \\
\Leftrightarrow \gamma_i &= \frac{{b_2}_i{c_1}_i - {b_1}_i{c_2}_i}{{c_1}_i + {b_2}_i - {b_1}_i - {c_2}_i}
\end{aligned}$$

It’s not pretty, but the important bits are there:

$$\begin{aligned}
\gamma_i &= \frac{{b_2}_i{c_1}_i - {b_1}_i{c_2}_i}{{c_1}_i + {b_2}_i - {b_1}_i - {c_2}_i} \\
\alpha &= \frac{{c_1}_i - {b_1}_i}{\gamma_i - {b_1}_i}
\end{aligned}$$

The good news: In theory, this totally works!

The bad news: In practice, it does not. This formula is *very* susceptible to rounding errors, and since we’re working with (and measuring) bytes, rather than floating-point numbers, rounding errors are a given, especially since the two background colors are so similar. Also, divisions by zero are possible, and, in fact, quite likely to occur.

<br>

Unless you can come up with an improvement to that idea this unfortunately looks like a dead end to me. Maybe we can instead just loop over all possible color values and find the closest match, instead? Given that there are only $2^{32}$ possible color combinations, doing so should not take more than perhaps a few minutes per color. It’s not exactly an elegant solution, but I cannot think of anything better at the moment, apart from maybe trying to solve this algorithmically in some way.

**EDIT:** I just had an idea. Rather than changing the window color by (de)activating wallpaper tinting, maybe we could just place the Swift components on top of an `NSBox` that is either fully black or fully white? Since the background colors would then differ so much, the rounding errors shouldn’t be as significant, then.

GroovinChip:
I don't really understand any of the math, but I'll try placing components onto fully opaque boxes and see what happens.

EDIT: OK so that was a really good idea. I'm gonna upload a video soon.

EDIT 2: Working on compressing the video so I can upload. Might not be able to continue till this evening.

GroovinChip:
@Adrian-Samoticha check this out

https://user-images.githubusercontent.com/4250470/225429145-ad1fa902-bb14-4f43-ba21-bb3ee7ed2e1c.mp4

Adrian-Samoticha:
I think that actually worked!

I ran the numbers and got the following color: `Color.fromRGBO(255, 255, 255, 0.11)`

After placing an object with that color on top of a black background in Inkscape I get this:

image

That’s the same exact color as the one in your video. Give me a second, I’ll write an automated script so you can use it too!

Adrian-Samoticha:
There you go:

```dart
void main() {
  // Change these:
  final b1 = <double>[0, 0, 0]; // The first background color.
  final c1 = <double>[28, 28, 28]; // The component color when rendered on top of b1.
  final b2 = <double>[255, 255, 255]; // The second background color.
  final c2 = <double>[255, 255, 255]; // The component color when rendered on top of b2.

  calculateAndPrint(b1, c1, b2, c2);
}

void calculateAndPrint(b1, c1, b2, c2) {
  final gamma = [0.0, 0.0, 0.0];
  final alpha = [0.0, 0.0, 0.0];
  for (int i = 0; i < 3; i += 1) {
    gamma[i] = (b2[i] * c1[i] - b1[i] * c2[i]) /
               (c1[i] + b2[i] - b1[i] - c2[i]);
    alpha[i] = (c1[i] - b1[i]) / (gamma[i] - b1[i]);
  }
  
  print('gamma: $gamma\nalpha: $alpha\n');
  
  print('Resulting color: Color.fromRGBO(${gamma[0].toInt()}, '
        '${gamma[1].toInt()}, ${gamma[2].toInt()}, '
        '${alpha.first.toStringAsFixed(3)})');
}
```

The code can be copy&pasted as-is and run in [DartPad](https://dartpad.dev/).
Just edit the color arrays in the `main` function and run it.

GroovinChip:
So I can give this function any RGB value and it will tell me the alpha value? In other words, if I inspect a native component and get an RGB value and give it to the function, it will tell me how much transparency is being applied to it?

Adrian-Samoticha:
You give it two background colors (b1 and b2) and two component colors the way they appear when rendered on top of their respective background colors (c1 and c2). In the video, I took the first black square (therefore, `final b1 = <double>[0, 0, 0];`) as well as the color of the switch that is rendered on top of it (therefore, `final c1 = <double>[28, 28, 28];`) and then took the second white square (therefore, `final b2 = <double>[255, 255, 255];`) as well as its component’s color (which is also white, therefore `final c2 = <double>[255, 255, 255];`).

The output is the color of the component, including its opacity (`Resulting color: Color.fromRGBO(255, 255, 255, 0.110)` in the case of the video).

GroovinChip:
Is the output the color for dark mode or light? Or is it the same for both?

Adrian-Samoticha:
That depends on whether dark mode is enabled for the component. In your video, you have dark mode enabled, so your switch is white. If you wanted to find out what color the switch should have in light mode you’d need to enable light mode and re-capture the colors.

Also, don’t forget to capture the color of the switch’s border, as that one is transparent as well, but has a different opacity.

GroovinChip:
Do you have any thoughts for figuring out what sort of effect is applied to buttons? Check this out:
image
Run the color meter over those buttons, watch how the value changes as you move your cursor. I have no idea what effect is being applied here. I thought it might have been shadows from button text, but with the text removed the effect remains.

Adrian-Samoticha:
Looks like a gradient to me. You can actually see it quite clearly when you zoom in (macOS has a Zoom feature hidden in its accessibility settings). In dark mode there is also a specular effect applied to the top of the button, which should be re-creatable using the [gradient_borders](https://pub.dev/packages/gradient_borders) package.

GroovinChip:
Man this is so complicated >.<

I'm going to publish my branch. If you're open to helping me figure this stuff out, I'd really appreciate it.

EDIT: branch name is called `accuracy_refactor`

GroovinChip:
I'm also having trouble getting the correct border color for the switch

Adrian-Samoticha:
Sure thing. By the way, where did you get that “native-components” window from? Is it a Swift project that is publicly available somewhere?

GroovinChip:
> Sure thing. By the way, where did you get that “native-components” window from? Is it a Swift project that is publicly available somewhere?

Thanks! 

So that app is a native SwiftUI project I created recently to assist my efforts to achieve better parity between this library and native apps. It's not very well made, I'm sure, and it's missing a lot. That said, I can put it up in a repo if you think it'll help.

Adrian-Samoticha:
Please do. Even if it isn’t perfect, it would still save me the trouble of creating it myself.

GroovinChip:
> Please do. Even if it isn’t perfect, it would still safe me the trouble of creating it myself.

[Here you go](https://github.com/macosui/native-components)

Adrian-Samoticha:
Alright, nothing final, yet, but I was able to achieve this:

image

Of course, it is not 100 % pixel-perfect. The secondary button’s shadow is a little bit off since I ran into https://github.com/flutter/flutter/issues/99344 while trying to offset it, but I don’t think there’s much I can do about it. I think I got very close, though.

All the colors are hard-coded, though, and implementing them into `MacosTheme` is going to be a little bit involved, since I am working with `BoxDecoration`s now, rather than just plain colors, but I’ll find a way.

That said, how do we proceed, now? Can I just merge `accuracy_refactor` into my own fork and continue working from there?

GroovinChip:
That looks great! Here's what I'm thinking right now: 

I want version 2.0 to be a very broad and truly major update. I haven't had intentions to make a 2.0 upgrade, unless there were such updates, because the library is stable, robust, and has lots of widgets. That was before I realized just how inaccurate most, if not all, of the widgets are. The design resources, both providing officially by Apple and unofficially in the figment kit, are good for getting most of the way there, but when held up against actual native components it is obvious that there is a lot of fidelity that doesn't translate to the designs. I'm not really comfortable releasing a major version change without including updates that address the missing fidelity. So I'm thinking that we can merge this PR into dev as a 2.0 alpha, and continue our efforts to address fidelity through the `accuracy_refactor` branch. We can merge updates into Dev as we progress through that effort.

What do you think?

Adrian-Samoticha:
Sounds good to me. Let me resolve the merge conflicts, first, though.

GroovinChip:
> Changes the default canvasColor to mimic [NSColor.windowBackgroundColor](https://developer.apple.com/documentation/appkit/nscolor/1528630-windowbackgroundcolor).

By the way - I discovered that through my accuracy research that setting `canvasColor` to `windowBackgroundColor` is actually wrong. By default, the background color of native apps is actually `underPageBackgroundColor`.

Adrian-Samoticha:
Are you sure? [`underPageBackgroundColor` is deprecated according to the Apple dev docs](https://developer.apple.com/documentation/uikit/uicolor/1623406-underpagebackgroundcolor?language=objc).

**EDIT:** Nevermind, that was a link to the UIKit docs, not to the AppKit ones.

**EDIT 2:** `underPageBackgroundColor` is apparently transparent in light mode, which is odd for a background color:

image

How did you conduct your research?

GroovinChip:
Well, I noticed that the RGB values shown for the background of a native app, which are 40, 40, 40, didn't match the values for `windowBackgroundColor`, which are 50, 50, 50. This also tracked with that I found in [this page](https://blog.verslu.is/xamarin/ios-macos-dark-mode-dynamic-colors-overview/), which walks through the RGB (and sometimes A) values for the colors. I'm not certain how accurate all the A values are, but in this case it seems to be correct when lining up a native app in light/dark mode with our example app.

GroovinChip:
Also, where'd you get that screenshot from? This one: 
image

Adrian-Samoticha:
It’s from the [appkit_ui_element_colors](https://github.com/Adrian-Samoticha/appkit_ui_element_colors) example project.

GroovinChip:
> It’s from the [appkit_ui_element_colors](https://github.com/Adrian-Samoticha/appkit_ui_element_colors) example project.

Neat!

Adrian-Samoticha:
You’re right about the actual background color being different from `windowBackgroundColor`, though, so maybe I’ll just capture some values from the Digital Color Meter and use those, instead.

Adrian-Samoticha:
Alright. As far as I’m concerned, this PR is ready to be merged.

GroovinChip:
> Alright. As far as I’m concerned, this PR is ready to be merged.

Cool. I'll pull it down, try it out, and review.

GroovinChip:
@Adrian-Samoticha I have to confess that I'm very overwhelmed by all the changes and I'm not certain where to start my review and how to move through them in a manner that reflects the hierarchy of the changes. Do you have any suggestions?

Adrian-Samoticha:
Sure. I think the most important thing to note is that macos_window_utils provides a widget called [`VisualEffectSubviewContainer`](https://pub.dev/documentation/macos_window_utils/latest/widgets_visual_effect_subview_container_visual_effect_subview_container/VisualEffectSubviewContainer/VisualEffectSubviewContainer.html) which applies a visual effect subview to its content. It is provided an [`NSVisualEffectViewMaterial`](https://pub.dev/documentation/macos_window_utils/latest/macos_ns_visual_effect_view_material/NSVisualEffectViewMaterial.html) which maps directly to [`NSVisualEffectView.Material`](https://developer.apple.com/documentation/appkit/nsvisualeffectview/material). This widget is used by the `WallpaperTintedArea` and [`TransparentMacOSSidebar`](https://pub.dev/documentation/macos_window_utils/latest/widgets_transparent_macos_sidebar/TransparentMacOSSidebar-class.html) widgets.

You will find that these widgets are used throughout macos_ui to style the window. In the `lib/src/layout/window.dart` file, the sidebar is wrapped with a `TransparentMacOSSidebar` widget to apply the `.sidebar` effect to it. Additionally, many surfaces are wrapped inside a `WallpaperTintedArea` to apply the `.windowBackground` effect to them.

To be precice, the wallpaper tinted area works by wrapping its content with a `VisualEffectSubviewContainer`, while also drawing its child using the [`BlendMode.src`](https://api.flutter.dev/flutter/dart-ui/BlendMode.html#src) blend mode. This is done to hide any widgets that may be underneath it and make it appear as if it was opaque, but in reality, it is transparent and only shows the visual effect subview underneath it.

There is one issue, though — Flutter’s [`BackDropFilter`](https://api.flutter.dev/flutter/widgets/BackdropFilter-class.html) class doesn't support transparency and implementing your own backdrop filter is currently impossible (see https://github.com/flutter/flutter/issues/16296). 

This makes it impossible use wallpaper tinted areas while displaying widgets that use a backdrop filter (such as context menus using the `MacosOverlayFilter` widget). The (hopefully temporary) workaround was to disable wallpaper tinting whenever an overlay filter is present in the widget tree. Disabling wallpaper tinting just means that we are drawing a background, instead. This is done by using the `GlobalWallpaperTintingSettings` class’ API to add or remove wallpaper titing overrides. The `WallpaperTintedArea` widget then uses a `WallpaperTintingSettingsBuilder` to detect if wallpaper tinting is enabled and build its widget subtree accordingly.

I tried to document all my classes and fields to the best of my ability, so if anything is unclear, be sure to check out the documentation.

GroovinChip:
Thank you!

pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: macos_ui
 description: Flutter widgets and themes implementing the current macOS design language.
-version: 1.12.3
+version: 2.0.0
```
GroovinChip:
```suggestion
version: 2.0.0-beta.1
```

lib/src/layout/wallpaper_tinted_area.dart
```diff
@@ -0,0 +1,141 @@
+import 'package:flutter/cupertino.dart';
+import 'package:macos_ui/src/layout/wallpaper_tinting_settings/global_wallpaper_tinting_settings.dart';
+import 'package:macos_ui/src/layout/wallpaper_tinting_settings/wallpaper_tinting_settings_builder.dart';
+import 'package:macos_window_utils/macos/ns_visual_effect_view_material.dart';
+import 'package:macos_window_utils/widgets/visual_effect_subview_container/visual_effect_subview_container.dart';
+
+class WallpaperTintedArea extends StatelessWidget {
```
GroovinChip, line 25: Let's add some top-level docs here

lib/src/layout/wallpaper_tinted_area.dart
```diff
@@ -0,0 +1,141 @@
...
+          ),
+          child: child,
+        );
+      },
+      child: Opacity(
+        // For some reason, omitting this Opacity widget causes
+        // a dark background to appear.
+        opacity: 1.0,
+        child: child,
+      ),
```
GroovinChip: Can you elaborate on this further? I think we ought to investigate this more before we merge.

lib/src/layout/window.dart
```diff
@@ -47,6 +45,33 @@ class MacosWindow extends StatefulWidget {
   /// A sidebar to display at the right of the window.
   final Sidebar? endSidebar;
 
+  /// Whether wallpaper tinting should be disabled.
+  ///
+  /// By default, macos_ui applies wallpaper tinting to the application's
```
GroovinChip:
```suggestion
  /// By default, `macos_ui` applies wallpaper tinting to the application's
```

lib/src/layout/window.dart
```diff
@@ -47,6 +45,33 @@ class MacosWindow extends StatefulWidget {
...
+  /// affected by wallpaper tinting transparent. Since Flutter's
+  /// [`ImageFilter.blur`](https://api.flutter.dev/flutter/dart-ui/ImageFilter/ImageFilter.blur.html)
+  /// does not support transparency, wallpaper tinting is disabled automatically
+  /// when a [MacosOverlayFilter] is present in the widget tree.
+  ///
+  /// This is meant to be a temporary solution until
+  /// [#16296](https://github.com/flutter/flutter/issues/16296) is resolved in
+  /// the Flutter project.
+  ///
+  /// Since the disabling of wallpaper tinting may be found to be too noticeable
```
GroovinChip:
```suggestion
  /// Since the disabling of wallpaper tinting may be found to be too noticeable, 
```

lib/src/layout/window.dart
```diff
@@ -47,6 +45,33 @@ class MacosWindow extends StatefulWidget {
...
+  /// this property may be used to disable wallpaper tinting outright.
+  final bool disableWallpaperTinting;
+
+  /// The [NSVisualEffectViewState] of the sidebar.
+  ///
+  /// Defaults to [NSVisualEffectViewState.followsWindowActiveState].
```
GroovinChip: Perhaps we ought to elaborate a bit here on what `NSVisualEffectViewState` is?

lib/src/layout/window.dart
```diff
@@ -139,26 +169,17 @@ class _MacosWindowState extends State<MacosWindow> {
...
-      // Only show blurry, transparent sidebar when platform brightness and app
-      // brightness are the same, otherwise it looks awful. Also only make the
-      // sidebar transparent on native Mac, or it will just be flat black or
-      // white.
-      sidebarBackgroundColor = Colors.transparent;
     } else {
-      sidebarBackgroundColor = theme.brightness.isDark
-          ? CupertinoColors.tertiarySystemBackground.darkColor
-          : CupertinoColors.systemGrey6.color;
+      sidebarBackgroundColor = Colors.transparent;
```
GroovinChip: I'd like to avoid as much `material`as possible. Let's use `MacosColors.transparent` instead.

example/lib/main.dart
```diff
@@ -13,7 +13,54 @@ import 'package:provider/provider.dart';
...
+
+  // Auto-hide toolbar and menubar in full-screen mode.
+  final options = NSAppPresentationOptions.from({
+    NSAppPresentationOption.fullScreen,
+    NSAppPresentationOption.autoHideToolbar,
+    NSAppPresentationOption.autoHideMenuBar,
+    NSAppPresentationOption.autoHideDock,
+  });
+  options.applyAsFullScreenPresentationOptions();
+}
```
GroovinChip: I think we ought to add this class, and any other similar classes, to `macos_ui` and export from there, rather than leaving it up to users to implement.

example/lib/main.dart
```diff
@@ -55,23 +102,25 @@ class _WidgetGalleryState extends State<WidgetGallery> {
...
+            CupertinoIcons.add,
+          ),
+        ),
+    (bool isVisible) => const DialogsPage(),
+    (bool isVisible) => const ToolbarPage(),
+    (bool isVisible) => SliverToolbarPage(
+          isVisible: isVisible,
+        ),
+    (bool isVisible) => const TabViewPage(),
+    (bool isVisible) => const SelectorsPage(),
```
GroovinChip, line 84: Can you please explain why it was necessary to make this change, as well as the change to the `IndexedStack`?

lib/src/layout/wallpaper_tinting_settings/wallpaper_tinting_settings_builder.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:flutter/widgets.dart';
+import 'package:macos_ui/src/layout/wallpaper_tinting_settings/global_wallpaper_tinting_settings.dart';
+
+import 'wallpaper_tinting_settings_data.dart';
+
+class WallpaperTintingSettingsBuilder extends StatelessWidget {
```
GroovinChip, line 24: Let's add top-level docs here

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## [2.0.0]
```
GroovinChip:
```suggestion
## [2.0.0-beta.1]
```

README.md
```diff
@@ -226,96 +226,68 @@ See the documentation for customizations and `ToolBar` examples.
...
+
+  // Auto-hide toolbar and menubar in full-screen mode.
+  final options = NSAppPresentationOptions.from({
+    NSAppPresentationOption.fullScreen,
+    NSAppPresentationOption.autoHideToolbar,
+    NSAppPresentationOption.autoHideMenuBar,
+    NSAppPresentationOption.autoHideDock,
+  });
+  options.applyAsFullScreenPresentationOptions();
+}
```
GroovinChip: Again, I think stuff like this should be provided by the library.

README.md
```diff
@@ -226,96 +226,68 @@ See the documentation for customizations and `ToolBar` examples.
...
-  }
+  runApp(const MacosUIGalleryApp());
 }
-
 ```
 
-See [this issue comment](https://github.com/flutter/flutter/issues/59969#issuecomment-916682559) for more details on the new look and explanations for how it works.
-
-Please note that if you are using a title bar (`TitleBar`) in your `MacosWindow`, you should set the `toolbarStyle` of NSWindow to `.expanded`, in order to properly align the close, minimize, zoom window buttons. In any other case, you should keep it as `.unified`. This must be set beforehand, i.e. it cannot be switched in runtime.
+Please note that if you are using a title bar (`TitleBar`) in your `MacosWindow`, you should set the `toolbarStyle` of NSWindow to `NSWindowToolbarStyle.expanded`, in order to properly align the close, minimize, zoom window buttons. In any other case, you should keep it as `NSWindowToolbarStyle.unified`.
```
GroovinChip: Let's clarify where the `toolbarStyle` is set. Maybe provide a code sample?

GroovinChip (changes requested):
@Adrian-Samoticha finally reviewed!!!

lib/src/layout/wallpaper_tinted_area.dart
Adrian-Samoticha: It’s been a while since I wrote this, but I believe the background was caused by the Cupertino scaffold. The `backgroundBlendMode: BlendMode.src` attribute of the `Container` inside the builder should, theoretically, remove that background, but for some reason, it doesn’t unless the `Opacity` widget is present. I’ll try to investigate further.

lib/src/layout/window.dart
Adrian-Samoticha: Perhaps. Though I find that the fact that it defaults to `NSVisualEffectViewState.followsWindowActiveState` makes it pretty self-explanatory. 

Even [Apple’s developer documentation](https://developer.apple.com/documentation/appkit/nsvisualeffectview/state) doesn’t go into much more detail than I do here.

README.md
Adrian-Samoticha: I chose to do it this way so that it was more faithful to how it was before the migration to macos_window_utils. Plus, I believe doing it this way may provide more customization options for the developer. If this functionality should be provided by the library, then I’d need to think about how to do that in a way that doesn’t sacrifice said customizability.

**EDIT:** I haven’t read your other comment, yet. Yes, I think exporting the classes from macos_ui does make sense.

example/lib/main.dart
Adrian-Samoticha: Admittedly, this isn’t a very elegant solution, but I just couldn’t find a better one. You see, `SliverToolbarPage` contains a transparent sliver that uses a blurry backdrop filter. Since blurry backdrop filters and wallpaper tinting don’t mix, I, of course, disable wallpaper tinting when the sliver is present in the widget tree. However, since the `SliverToolbarPage` is hidden inside an `IndexedStack`, it is not always visible, so I don’t need to always disable wallpaper tinting — only when the toolbar actually visible on-screen.

For this reason, I needed a way to determine if the `SliverToolbarPage` was currently visible, which I did by replacing `pages` with a list of builders that can be informed if the widget they build is currently visible.

Had I not done this, wallpaper tinting would always be disabled since the `SliverToolbarPage` is always present in the widget tree.

The solution is supposed to be temporary and be reverted once custom backdrop filters are supported in Flutter, but until then, the only other solution I can think of is to use [flutter_bloc](https://pub.dev/packages/flutter_bloc) to manage the index’ state.

Do you think that would be a cleaner solution?

lib/src/layout/window.dart
GroovinChip: Alright.

> Even Apple’s developer documentation doesn’t go into much more detail than I do here.

I wouldn't call Apple's documentation something to live up to 😅

README.md
GroovinChip: Yes, I think we ought to provide classes that represent all possible use cases a developer could want.

example/lib/main.dart
GroovinChip: Fair enough.

No, I don't think we need to use bloc for this. We can leave it.

lib/src/layout/wallpaper_tinted_area.dart
Adrian-Samoticha: Turns out that the opacity widget could be replaced with a repaint boundary. I assume `Opacity` just used a `RepaintBoundary` under the hood which allowed the `BlendMode.src` to work correctly.

Adrian-Samoticha:
Alright. I think I addressed all points in your review.

example/lib/main.dart
```diff
@@ -13,7 +13,15 @@ import 'package:provider/provider.dart';
...
 import 'theme.dart';
 
-void main() {
+/// This method initializes macos_window_utils and styles the window.
+Future<void> _configureMacosWindowUtils() async {
+  const config = MacosWindowUtilsConfig();
+  await config.apply();
+}
+
+void main() async {
```
GroovinChip:
```suggestion
Future<void> main() async {
```

lib/src/macos_window_utils_config.dart
```diff
@@ -0,0 +1,120 @@
...
+  final bool doMakeTitlebarTransparent;
+
+  /// Whether to hide the title.
+  final bool doHideTitle;
+
+  /// Whether to remove the menubar in full-screen mode.
+  final bool doRemoveMenubarInFullScreenMode;
+
+  /// Whether to automatically hide the toolbar and menubar in full-screen mode.
+  final bool doAutoHideToolbarAndMenuBarInFullScreenMode;
```
GroovinChip: Can we rename these to drop the "do"? For example:
* `enableFullSizeContentView`
* `hideTitle`

GroovinChip (changes requested):
Looking good, let's just make these small changes.

lib/src/macos_window_utils_config.dart
Adrian-Samoticha: Done.

GroovinChip (approved):
LGTM 🎉 🎉 🎉

PR #414, opened 2023-04-16
feat: Added support for `startWeekOnMonday` to `MacosDatePicker`

rklos:
This is the first step in fulfilling all requests from https://github.com/macosui/macos_ui/issues/369.
I've added an optional parameter `startWeekOnMonday` which allows to render a calendar with Monday as a first day of the week. 

This is a temporary solution, because the need of this feature is caused by invalid implementation of `firstDayOfWeekIndex` in `flutter_localizations` package: https://github.com/flutter/flutter/issues/122274

## Pre-launch Checklist

- [X] I have incremented the package version as appropriate and updated `CHANGELOG.md` with my changes 
- [X] I have added/updated relevant documentation 
- [X] I have run "optimize/organize imports" on all changed files
- [X] I have addressed all analyzer warnings as best I could

PR #415, opened 2023-04-16
feat: Added support for `dateFormat` to `MacosDatePicker`

rklos:
This is another feature from https://github.com/macosui/macos_ui/issues/369.
I've added an optional parameter `dateFormat` which allows to change the way dates are displayed in the textual interface.

## Pre-launch Checklist

- [X] I have incremented the package version as appropriate and updated `CHANGELOG.md` with my changes 
- [X] I have added/updated relevant documentation 
- [X] I have run "optimize/organize imports" on all changed files
- [X] I have addressed all analyzer warnings as best I could

### DIFF
--- a/lib/src/selectors/date_picker.dart
+++ b/lib/src/selectors/date_picker.dart
@@ -44,6 +44,7 @@ class MacosDatePicker extends StatefulWidget {
     this.style = DatePickerStyle.combined,
     required this.onDateChanged,
     this.initialDate,
+    this.dateFormat,
   });
 
   /// The [DatePickerStyle] to use.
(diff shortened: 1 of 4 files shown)


lib/src/selectors/date_picker.dart
```diff
@@ -59,6 +60,8 @@ class MacosDatePicker extends StatefulWidget {
   /// Defaults to `DateTime.now()`.
   final DateTime? initialDate;
 
+  final String? dateFormat;
```
GroovinChip, line 75: Please add some documentation for this.

GroovinChip (changes requested):
Looking good, just merge in the latest from `dev`, adjust the beta version where appropriate, and make the requested change. Thanks!

lib/src/selectors/date_picker.dart
rklos: Done

lib/src/selectors/date_picker.dart
```diff
@@ -59,6 +60,18 @@ class MacosDatePicker extends StatefulWidget {
   /// Defaults to `DateTime.now()`.
   final DateTime? initialDate;
 
+  /// Changes the way dates are displayed in the textual interface.
+  /// The following tokens are supported (case-insensitive):
```
GroovinChip:
```suggestion
  /// Changes the way dates are displayed in the textual interface.
  /// 
  /// The following tokens are supported (case-insensitive):
```

GroovinChip (changes requested):
Just a minor dartdoc formatting change needed, then we're good to go.

GroovinChip (approved):
LGTM. Waiting on checks and #414.

Issue #271, opened 2022-06-30
Sidebar should support n levels of nesting to match MacOS UI


slewallen:
**Description**

The Sidebar on Mac OS, such as in Mail, allows a full tree UI with n levels of nested nodes.  The macos_ui Sidebar is only supporting one level of nesting.

**Steps To Reproduce**

Add the following snippet to the macos_ui widget gallery:

                SidebarItem(
                  label: Text('Level 1 Group'),
                  leading: MacosIcon(FontAwesomeIcons.layerGroup),
                  disclosureItems: [
                    SidebarItem(
                      leading: MacosIcon(FontAwesomeIcons.grid),
                      label: Text('Layer 1'),
                    ),
                    SidebarItem(
                      leading: MacosIcon(FontAwesomeIcons.grid),
                      label: Text('Layer 2'),
                    ),
                    SidebarItem(
                      label: Text('Level 1b Group'),
                      leading: MacosIcon(FontAwesomeIcons.layerGroup),
                      disclosureItems: [
                        SidebarItem(
                          leading: MacosIcon(FontAwesomeIcons.grid),
                          label: Text('Object 1'),
                        ),
                        SidebarItem(
                          leading: MacosIcon(FontAwesomeIcons.grid),
                          label: Text('Object 2'),
                        ),
                      ],
                    ),

The result is that the 'Level 1b Group' has no disclosure turn-down and no children. 

<img width="190" alt="NodeNesting" src="https://user-images.githubusercontent.com/23299259/176790461-25c9c4ed-1a37-415b-833e-92225d532edb.png">

Mail on MacOS, as an example, does support n number of nested nodes.

<img width="274" alt="MailNestedFolders" src="https://user-images.githubusercontent.com/23299259/176790607-9674d13a-56ef-485d-a2ec-8f68b49aeb02.png">

GroovinChip:
@slewallen Did you mean to add more information to this issue? It appears to be empty.

GroovinChip:
This is an interesting one...I'm guessing it has to do with how we render the disclosure items:
```dart
children: widget.item.disclosureItems!.map((item) {
  return Padding(
    padding: EdgeInsets.only(
      left: 24.0 + theme.visualDensity.horizontal,
    ),
    child: SizedBox(
      width: double.infinity,
      child: _SidebarItem(
        item: item,
        onClick: () => widget.onChanged?.call(item),
        selected: widget.selectedItem == item,
      ),
    ),
  );
}).toList(),
```
[...]

PR #465, opened 2023-07-26
Enhancement - new hasActiveSelection property for SidebarItems

ganeshrvel:
### 🔄 Updated 🔄
* `SidebarItems` now accepts `hasActiveSelection` property which indicates whether any of the items in the sidebar has been selected.

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [2.0.1]
+### 🔄 Updated 🔄
+* `SidebarItems` now accepts `hasActiveSelection` property which indicates whether any of the items in the sidebar has been selected.
+
 ## [2.0.0]
 ### 🚨 Breaking Changes 🚨
 * `macos_ui` has been migrated to utilize [macos_window_utils](https://pub.dev/packages/macos_window_utils) under the hood, which provides the following benefits:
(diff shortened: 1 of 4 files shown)


GroovinChip:
Hi @ganeshrvel, please post a screenshot of your change

ganeshrvel:
Screen recording of the changes:

https://github.com/macosui/macos_ui/assets/182092/49e8d1cd-ad28-414b-b63f-f330265c7fd2

GroovinChip:
Thanks for the recording. What's the purpose of this change?

ganeshrvel:
> Thanks for the recording. What's the purpose of this change?

Currently, the SidebarItems are built with the presumption that any selection within the content area should occur as a result of an action on one of the SidebarItems.

However, a sidebar could contain multiple items, including multiple SidebarItems, Texts, Buttons, etc., and the displayed contents may not necessarily be directly related to the items on the Sidebar.

For example, consider the File explorer sidebar on macOS or Windows. Users may navigate between files, and only when they open a pinned item on the side, it should be highlighted.

ganeshrvel:
@GroovinChip bumping this reply as a reminder

GroovinChip:
> @GroovinChip bumping this reply as a reminder

Little busy, mate. I'll get to it.

GroovinChip:
@ganeshrvel so the thing is, we need to implement #271 and I'm reluctant to accept any changes related to the sidebar until we do. Do you think your change might collide with such work? Or would you perhaps be willing to take #271 on and fold this change into it, if it's even still needed at that point?

GroovinChip:
I am closing this due to inactivity.

PR #467, opened 2023-07-31
Alter copy of mock_canvas to unblock moving to flutter_test

Piinks:
Came up as a test failure in https://github.com/flutter/flutter/pull/131631

Since mock_canvas was copied into this repo, the analysis will fail when we try to make it public. This just renames the methods so we can land the change. Once it is part of flutter_test, the copy of recording_canvas and mock_canvas can be removed entirely since it will be public.

The commit in flutter/tests will need to be updated after this for the PR in flutter/flutter to be able to land.

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [2.0.1]
+* Modified mock_canvas copy from the framework to unblock move to flutter_test
+
 ## [2.0.0]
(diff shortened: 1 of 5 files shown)


Piinks:
@GroovinChip

GroovinChip:
Amazing! Thank you, Kate!

GroovinChip (approved):
LGTM!

PR #469, opened 2023-08-02
Expose NSWindow Callbacks for Enhanced Window Handling

ganeshrvel:
This PR proposes exposing several NSWindow event callbacks, such as `onWindowDidBecomeMain`, `onWindowDidResignMain`, `onWindowDidBecomeKey`, `onWindowDidResignKey`, `onWindowDidChangeScreen`, `onWindowDidMove`, and `onWindowDidResize`, to allow developers to leverage and customize window behavior better.

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## [2.0.3]
+### ✨ New ✨
+- Exposed the following NSWindow event callbacks via input parameters:
+  - `onWindowDidBecomeMain`
+  - `onWindowDidResignMain`
+  - `onWindowDidBecomeKey`
+  - `onWindowDidResignKey`
+  - `onWindowDidChangeScreen`
+  - `onWindowDidMove`
+  - `onWindowDidResize`
+
 ## [2.0.0]
 ### 🚨 Breaking Changes 🚨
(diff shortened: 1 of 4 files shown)


GroovinChip:
@Adrian-Samoticha what are your thoughts on this? Do we want to route more of MWU through MUI, or would you rather folks use MWU directly?

Adrian-Samoticha:
> @Adrian-Samoticha what are your thoughts on this? Do we want to route more of MWU through MUI, or would you rather folks use MWU directly?

Frankly, I do like this proposal. I think since we’re already routing MWU calls through MUI it probably doesn’t hurt to expand on that. However, I’d argue that we should go all the way and expose *all* `NSWindowDelegate` callbacks in that case. 

Furthermore, I’d argue that using `VoidCallback` here isn’t ideal, since that means that you can only define a single function for each callback. I’d suggest we instead define a sort of “window event stream” which emits an enum corresponding to each `NSWindowDelegate` method whenever an event is triggered.

GroovinChip:
I agree @Adrian-Samoticha. @ganeshrvel, can you please make the proposed changes?

GroovinChip:
@ganeshrvel I am closing this due to inactivity. Please feel free to reopen when you are ready to make the requested changes.

PR #471, opened 2023-08-07
Resolve #445

### LINKED ISSUE #445: `PushButton` fidelity
`PushButton` requires several fidelity improvements for parity with native push buttons

```[tasklist]
### Tasks
- [ ] Determine and apply the appropriate gradients, box shadows, padding, and border radii for all control sizes
- [ ] Determine and apply correct press animation
- [ ] Apply correct secondary background colors
- [ ] Apply correct disabled background colors
```

Adrian-Samoticha:
This PR resolves #445.

`PushButton` now more closely mimics the look and feel of native macOS buttons:

Screenshot 2023-08-07 at 15 41 01

In addition to featuring gradients and specular highlights, it also becomes greyed out when the window loses its focus …

Screenshot 2023-08-07 at 15 42 44

… and correctly responds to the user’s preferred accent color:

image

It does currently **not** respect the user’s choice of the “Increase contrast” setting in the Accessibility Settings. I’d suggest tackling that point in a future issue.

## Technical Details

The `utils.dart` file now contains a `WindowMainStateListener` class, allowing listening to focus changes in the window. ~It currently has a minor issue: It assumes that the window is focused during startup and doesn’t check if it actually is. This is likely true in a released application but may be incorrect during testing (since the app can be restarted from within VS Code or the terminal). This issue will be fixed once https://github.com/macosui/macos_window_utils.dart/issues/31 lands.~ **EDIT:** This is now fixed.

Additionally, there now exists an `AccentColorListener` class. It is used by `PushButton` to retrieve the current accent color and listen to changes performed by the user. 

The push-button’s `pressedOpacity` property and `PushButtonTheme` have also been deprecated, however, not all references to `PushButtonTheme` have been removed from the `MacosThemeData` class to ensure backward compatibility. I suppose they should be removed in the next breaking release, though.

## Pre-launch Checklist


macwilko:
Amazing work!

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [2.0.1]
+* `PushButton` has received a facelift. It now mimics the look of native macOS buttons more closely.
```
GroovinChip:
```suggestion
* `PushButton` has received a facelift. It now mimics the look and feel of native macOS buttons more closely.
```

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
     _lastBrightness = currentBrightness;
   }
 }
+
+/// A class that listens for changes to the application's window being the main
+/// window, and notifies listeners.
+class WindowMainStateListener {
```
GroovinChip: I think this probably ought to be either: 
1. placed under the `layout` directory, since it's directly related
2. placed in its own file in a dedicated `utils` directory

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
     _lastBrightness = currentBrightness;
   }
 }
+
+/// A class that listens for changes to the application's window being the main
+/// window, and notifies listeners.
```
GroovinChip:
```suggestion
/// A class that listens for changes to the application's main window. 
///
/// A common use-case for responding to such changes would be to mute the colors of certain
/// primary UI elements when the window is no longer in focus, which is something native
/// macOS applications do out of the box.
```

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+  static final instance = WindowMainStateListener();
+
+  /// A [NSWindowDelegateHandle], to be used when deiniting the listener.
+  NSWindowDelegateHandle? handle;
+
+  /// Whether the window is currently the main window.
+  bool _isMainWindow = true;
+
+  /// Whether the window is currently the main window.
+  bool get isMainWindow => _isMainWindow;
```
GroovinChip: Do we need a private field and a public getter? In Dart, all instance variables have [implicit getters](https://dart.dev/language/methods#getters-and-setters) and setters.

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+  /// Initializes the [_isMainWindow] variable.
+  Future<void> _initIsWindowMain() async {
+    _isMainWindow = await WindowManipulator.isMainWindow();
+    _windowMainStateStreamController.add(_isMainWindow);
+  }
+
+  /// Deinitializes the listener.
+  void deinit() {
+    handle?.removeFromHandler();
+  }
```
GroovinChip: Should this not be called dispose, per general Flutter convention?

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+  final void Function() onWindowDidResignMain;
+
+  @override
+  void windowDidBecomeMain() => onWindowDidBecomeMain();
+
+  @override
+  void windowDidResignMain() => onWindowDidResignMain();
+}
+
+/// A class that listens to accent color changes.
```
GroovinChip:
```suggestion
/// A class that listens to changes to the user's selected system accent color.
///
/// Native macOS applications respond to such changes immediately.
```
We should also add a sample code snippet demonstrating usage, unless this is something that only happens under the hood.

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+    0.11813830353929083: AccentColor.yellow,
+    0.29428158007138466: AccentColor.green,
+    0.0: AccentColor.graphite,
+  };
+
+  /// The currently active accent color.
+  AccentColor? _currentAccentColor;
+
+  /// The currently active accent color.
+  AccentColor? get currentAccentColor => _currentAccentColor;
```
GroovinChip: Same question as earlier regarding private instance field with public getter, as this isn't really necessary in Dart

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+
+  /// Notifies listeners when the accent color changes.
+  final _accentColorStreamController = StreamController<void>.broadcast();
+
+  /// An accent color stream. Emits a new value whenever the accent color
+  /// changes.
```
GroovinChip:
```suggestion
  /// Streams the user's system accent color selection. 
  ///
  /// Emits a new value whenever the system accent color selection changes.
```

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+    if (!Platform.isMacOS) return;
+
+    _initCurrentAccentColor();
+    _initSystemColorObserver();
+  }
+
+  /// Deinitializes this class.
+  void deinit() {
+    _systemColorObserverStreamSubscription?.cancel();
+  }
```
GroovinChip: Same question about "deinit" versus "dispose"

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+      return hueComponentToAccentColor[hueComponent]!;
+    }
+
+    debugPrint(
+      'Warning: Falling back on slow accent color resolution. It’s possible '
+      'that the accent colors have changed in a recent version of macOS, thus '
+      'invalidating macos_ui’s accent colors, which were captured in macOS '
+      'Ventura. If you see this message, please notify a maintainer of the '
+      'macos_ui package.',
+    );
```
GroovinChip: I like this!

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+
+    return _slowlyResolveAccentColorFromHueComponent(hueComponent);
+  }
+
+  /// This is a fallback method in case the above method fails.
```
GroovinChip:
```suggestion
  /// This is a fallback method in case [_resolveAccentColorFromHueComponent] fails.
```

pubspec.yaml
```diff
@@ -11,7 +11,9 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  macos_window_utils: ^1.1.3
+  macos_window_utils: ^1.2.0
+  gradient_borders: ^1.0.0
+  appkit_ui_element_colors: ^1.0.0
```
GroovinChip, line 16: I thought you weren't really confident in the values returned by this plugin?

lib/src/enums/accent_color.dart
```diff
@@ -0,0 +1,27 @@
+/// The macOS accent color which can be changed by the user in *System Settings*
+/// → *Appearance* → *Accent color*.
+enum AccentColor {
```
GroovinChip, line 3: Can this enum also be used to manually retrieve these colors? Basically, does it replace the corresponding color values in the `MacosColors` class?

lib/src/utils.dart
```diff
@@ -86,3 +90,219 @@ class MacOSBrightnessOverrideHandler {
...
+  AccentColor _slowlyResolveAccentColorFromHueComponent(double hueComponent) {
+    final entries = hueComponentToAccentColor.entries;
+    var lowestDistance = double.maxFinite;
+    var toBeReturnedAccentColor = AccentColor.values.first;
+
+    for (var entry in entries) {
```
GroovinChip:
```suggestion
    for (final entry in entries) {
```

lib/src/buttons/push_button.dart
```diff
@@ -169,6 +171,8 @@ class PushButton extends StatefulWidget {
   ///
   /// This defaults to 0.4. If null, opacity will not change on pressed if using
   /// your own custom effects is desired.
+  @Deprecated("'PushButton' now attempts to mimic macOS’ look and feel, "
+      "therefore, its opacity no longer changes when it is pressed.")
```
GroovinChip:
```suggestion
  @Deprecated("'PushButton' animations now match their native macOS’ counterparts. "
      "Therefore, its opacity no longer changes when it is pressed.")
```

GroovinChip (changes requested):

lib/src/utils.dart
Adrian-Samoticha: I wanted to encapsulate that field so that it couldn’t be set from outside the class, while still remaining gettable.

lib/src/utils.dart
Adrian-Samoticha: Not sure, but it’s not a widget, after all.

lib/src/utils.dart
Adrian-Samoticha: It’s technically meant to only be used under the hood, although exposing it to macos_ui users probably wouldn’t really hurt, either. In both cases, adding code snippets is a good idea, even if they’re meant to only be read only by contributors, so I’ll do that.

lib/src/utils.dart
GroovinChip: Fair enough!

pubspec.yaml
Adrian-Samoticha: I only use it to figure out the currently selected system accent color. The way I did it is by capturing the hue of a color that I know is dependent on the currently selected accent color and matching it against previously captured hue values. While I do indeed not know if the colors it provides are useful for rendering purposes, it definitely works for this purpose.

lib/src/enums/accent_color.dart
Adrian-Samoticha: I don’t think so. The enum maps to accent color *settings*, rather than actual color values.

lib/src/utils.dart
Adrian-Samoticha: I was under the impression that the `layout` directory was dedicated to widgets. I guess adding a `utils` directory would indeed make more sense, though.

**EDIT:** If I were to move those classes into a dedicated `utils` directory, should I move the `utils.dart` file there also?

GroovinChip: Yes, I'd say so. And perhaps what's in there should be better organized.,

lib/src/utils.dart
GroovinChip: Do we know if there is a convention for Dart-but-not-Flutter closing of resources? I'll ask some colleagues.

GroovinChip: So dispose is actually the convention in Dart as well, apparently

Adrian-Samoticha: Seems you might be right. I just looked through dart.dev and found [this](https://dart.dev/tools/linter-rules/cancel_subscriptions), so I’ll rename both `deinit` methods to `dispose`, then.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [2.0.1]
+* `PushButton` has received a facelift. It now mimics the look and feel of native macOS buttons more closely.
```
GroovinChip, line 3: Can we mark this with an "updated" like some previous sections have?

lib/src/buttons/push_button.dart
```diff
@@ -339,33 +340,363 @@ class PushButtonState extends State<PushButton>
...
+              MacosColor.fromRGBO(255, 255, 255, 0.251 * isEnabledFactor),
+            ]
+          : [
+              MacosColor.fromRGBO(255, 255, 255, 1.0 * isEnabledFactor),
+              MacosColor.fromRGBO(255, 255, 255, 1.0 * isEnabledFactor),
+            ];
+    }
+
+    if (isDarkModeEnabled) {
+      switch (accentColor) {
```
GroovinChip, line 414: I have to wonder if Dart 3 switch expressions would make this code a little less verbose? This is just a comment, not a change request.

lib/src/buttons/push_button.dart
```diff
@@ -339,33 +340,363 @@ class PushButtonState extends State<PushButton>
...
+        BoxShadow(
+          color: MacosColor.fromRGBO(0, 0, 0, 0.4 * isEnabledFactor),
+          blurRadius: 0.5,
+          offset: isEnabled ? const Offset(0.0, 0.3) : Offset.zero,
+          spreadRadius: 0.0,
+          blurStyle: isEnabled ? BlurStyle.normal : BlurStyle.outer,
+        ),
+      ];
+    } else {
+      switch (accentColor) {
```
GroovinChip, line 565: Similar thought about Dart 3 switch expressions (also not a change request)

lib/src/utils/accent_color_listener.dart
```diff
@@ -0,0 +1,160 @@
+import 'dart:async';
+import 'dart:io';
+import 'dart:math';
+
+import 'package:appkit_ui_element_colors/appkit_ui_element_colors.dart';
+import 'package:flutter/foundation.dart';
+import 'package:macos_ui/src/enums/accent_color.dart';
+
+/// A class that listens to changes to the user’s selected system accent color.
```
GroovinChip:
```suggestion
/// A class that listens for changes to the user’s selected system accent color.
```

lib/src/utils/accent_color_listener.dart
```diff
@@ -0,0 +1,160 @@
...
+  /// The currently active accent color.
+  AccentColor? get currentAccentColor => _currentAccentColor;
+
+  /// Notifies listeners when the accent color changes.
+  final _accentColorStreamController = StreamController<void>.broadcast();
+
+  /// Streams the user’s system accent color selection.
+  ///
+  /// Emits a new value whenever the system accent color selection changes.
+  Stream<void> get onChangedStream => _accentColorStreamController.stream;
```
GroovinChip: Does it makes sense to call this "onChangedStream" over "onChanged"? I think in practice when using it with a StreamBuilder, like in the docs example, it's a little redundant. We know it's a stream. I don't know; being straightforward is good, but I also lean towards brevity. Not a change request, just something I'm thinking about while reading this over.

lib/src/utils/accent_color_listener.dart
```diff
@@ -0,0 +1,160 @@
...
+    0.8285987697113538: AccentColor.purple,
+    0.9209523937489168: AccentColor.pink,
+    0.9861913496946438: AccentColor.red,
+    0.06543037411201169: AccentColor.orange,
+    0.11813830353929083: AccentColor.yellow,
+    0.29428158007138466: AccentColor.green,
+    0.0: AccentColor.graphite,
+  };
+
+  /// The currently active accent color.
```
GroovinChip:
```suggestion
  /// The active accent color selection.
```
"Active" already implies currency

lib/src/utils/accent_color_listener.dart
```diff
@@ -0,0 +1,160 @@
...
+    assert(_systemColorObserverStreamSubscription == null);
+
+    _systemColorObserverStreamSubscription =
+        AppkitUiElementColors.systemColorObserver.stream.listen((_) {
+      _initCurrentAccentColor();
+      _accentColorStreamController.add(null);
+    });
+  }
+
+  /// Returns the hue component of the currently active accent color on macOS.
```
GroovinChip:
```suggestion
  /// Returns the hue component of the active accent color selection on macOS.
```
"Active" already implies currency

lib/src/utils/accent_color_listener.dart
```diff
@@ -0,0 +1,160 @@
...
+  /// Returns the distance between two hue components.
+  double _distanceBetweenHueComponents(double component1, double component2) {
+    final rawDifference = (component1 - component2).abs();
+    return sin(rawDifference * pi);
+  }
+
+  /// A class that listens to accent color changes.
+  AccentColorListener() {
+    _init();
+  }
```
GroovinChip: Constructors should be sorted first, let's move this to the top of the class.

lib/src/utils/window_main_state_listener.dart
```diff
@@ -0,0 +1,109 @@
...
+///           = WindowMainStateListener.instance.isMainWindow;
+///
+///       return SomeWidget(
+///         isMainWindow: isMainWindow,
+///         child: …
+///       );
+///     },
+///  );
+/// ```
+class WindowMainStateListener {
```
GroovinChip, line 29: It occurs to me that with the eventuality of official multi-window support that calling this a window "main" state listener probably won't be correct; `WindowStateListener` will probably be more correct. Not an immediate issue, but if we correct for it now we'll have less work to do in the future.

lib/src/utils/window_main_state_listener.dart
```diff
@@ -0,0 +1,109 @@
...
+
+  /// Whether the window is currently the main window.
+  bool get isMainWindow => _isMainWindow;
+
+  /// Notifies listeners when the window’s main state changes.
+  final _windowMainStateStreamController = StreamController<bool>.broadcast();
+
+  /// A stream of the window’s main state. Emits a new value whenever the state
+  /// changes.
+  Stream<bool> get onChangedStream => _windowMainStateStreamController.stream;
```
GroovinChip: Could simply be called `onChanged` to avoid redundancy, but not a hard request.

lib/src/utils/window_main_state_listener.dart
```diff
@@ -0,0 +1,109 @@
...
+  final void Function() onWindowDidBecomeMain;
+
+  /// Called when the window resigns as the main window.
+  final void Function() onWindowDidResignMain;
+
+  @override
+  void windowDidBecomeMain() => onWindowDidBecomeMain();
+
+  @override
+  void windowDidResignMain() => onWindowDidResignMain();
```
GroovinChip, line 108: Now _these_ names make sense in the context of the eventual multi-window framework support :)

GroovinChip (changes requested):

lib/src/utils/accent_color_listener.dart
Adrian-Samoticha: I guess omitting the `Stream` does make sense. [I’ve looked up how other people name their streams on grep.app](https://grep.app/search?current=2&q=child%3A%20StreamBuilder%3C) and it does seem that omitting the stream is common.

lib/src/utils/window_main_state_listener.dart
Adrian-Samoticha: I mean, it will still listen to the windows’ main state, it will just need to be updated such that it can handle multiple windows, so I think the name is fine. Admittedly, I don’t think Apple actually refers to the window being the main window as its “state”, but I don’t think that’s inherently incorrect either.

GroovinChip: When sub windows are in focus, are they considered the main window? I guess that would be the determining factor.

Adrian-Samoticha: I’m fairly certain that the focused window is the main one.

**EDIT:** Admittedly, [Apple’s documentation](https://developer.apple.com/documentation/appkit/nswindow/1419130-ismainwindow) may be a little misleading here. They say

> A Boolean value that indicates whether the window is the application’s main window.

, which kinda sounds as if it is referring to the main window of the application, rather than the currently focused one, but still. Considering that a window resigns its main status when you select a different window I’m pretty sure that the main window is the one that’s currently in focus.

GroovinChip: Then I suppose we need to determine what a window's "main" state is

lib/src/buttons/push_button.dart
Adrian-Samoticha: That’d be cool. I wasn’t a fan of the level of nesting in this method, either.

GroovinChip (approved):
LGTM

Issue #359, opened 2023-02-18
Give the ResizablePane the option to add a toolbar


FelixMethe:
## Use case
This would make it possible to add different toolbars to the `ResizablePane` and `ContentArea` like the mail app does for example.
<img width="741" alt="image" src="https://user-images.githubusercontent.com/42588649/219856638-34628ce6-3cd7-48cf-b542-cabfc6dcc864.png">

This is not possible now because the `ResizablePane` adds a `MacosScrollbar` over the content of the pane.

## Proposal
I propose to add an optional toolbar argument to the `ResizablePane` that adds the toolbar above the content of the pane.

Please let me know if this is a desired feature.

PR #487, opened 2023-10-15
chore: allow customising window handle safe area padding

junaid1460:
Situation: We should be able to use toolbar within sub view and that need not have this padding 

Solution: Let end dev customise this padding

## Pre-launch Checklist

- [ ] I have addressed all analyzer warnings as best I could


junaid1460:
Still a draft, I will write tests and let you know when it's ready.

lib/src/layout/toolbar/sliver_toolbar.dart
```diff
@@ -41,6 +41,7 @@ class SliverToolBar extends StatefulWidget with Diagnosticable {
     this.floating = false,
     this.toolbarOpacity = 0.9,
     this.allowWallpaperTintingOverrides = true,
+    this.windowHandlePadding,
```
junaid1460, line 44: safeAreaPadding?

GroovinChip:
@junaid1460 I assume this aims to address #359?

junaid1460:
> @junaid1460 I assume this aims to address #359?

No, I didn't look at it before this, I just tried this library and found this padding blocking my design goal, it was hardcoded.
Great if it solves that too. from what I read, it should.

junaid1460:
I have made couple of more changes. un related to original change, pulldown button isn't customizable right now

GroovinChip:
@junaid1460 are you still working on this?

junaid1460:
> @junaid1460 are you still working on this?

Nope, I think we can close it or if you like to take over please do so. I have lost the context.

PR #526, opened 2024-10-30
Add icon support to push buttons (#522)

### LINKED ISSUE #522: Push Buttons with Icons
## Use case

I know that the current proposed solution to this problem is to use a MacosIconButton but the look this gives is not the same as a push button.  In native SwiftUI, you can easily create a primary or (more commonly) a secondary push button with an icon in it.

## Proposal

Expand the MacosIconButton class to allow you to style it like a push button with an easy toggle (something like how the PushButton class allows you to turn it into a secondary button by setting the "secondary" boolean property) OR simply allow passing an icon to the push button class to make it an icon button instead.

skythomp16:
This PR adds support for adding an icon to a push button.  This approach allows passing in an iconData instead of an icon so that we can control the size of the icon proportional to the button size.  I have also updated the buttons_page.dart class for the example to include both primary and secondary buttons with icons.  Comparing against a fresh SwiftUI project side by side, these look identical to those produced by it.  

## Pre-launch Checklist

### DIFF
--- a/lib/src/buttons/push_button.dart
+++ b/lib/src/buttons/push_button.dart
@@ -77,6 +77,34 @@ extension PushButtonControlSizeX on ControlSize {
     }
   }
 
+  /// Determines the button's icon size
+  double get iconSize {
+    switch (this) {
+      case ControlSize.mini:
+        return 8;
+      case ControlSize.small:
+        return 12;
+      case ControlSize.regular:
+        return 16;
+      case ControlSize.large:
+        return 20;
+    }
+  }
+
+  /// Determines the button's left icon padding
+  EdgeInsets get iconPadding {
+    switch (this) {
+      case ControlSize.mini:
+        return const EdgeInsets.fromLTRB(5, 0, 0, 0);
+      case ControlSize.small:
+        return const EdgeInsets.fromLTRB(7, 0, 0, 0);
+      case ControlSize.regular:
+        return const EdgeInsets.fromLTRB(10, 0, 0, 0);
+      case ControlSize.large:
+        return const EdgeInsets.fromLTRB(12, 0, 0, 0);
+    }
+  }
+
   /// Determines the button's minimum size.
   BoxConstraints get constraints {
     switch (this) {
@@ -130,6 +158,7 @@ class PushButton extends StatefulWidget {
     this.semanticLabel,
     this.mouseCursor = SystemMouseCursors.basic,
     this.secondary,
+    this.iconData,
   }) : assert(pressedOpacity == null ||
             (pressedOpacity >= 0.0 && pressedOpacity <= 1.0));
 
@@ -202,6 +231,10 @@ class PushButton extends StatefulWidget {
   /// Can still be overridden if the [color] attribute is non-null.
   final bool? secondary;
 
+  /// An optional iconData that can be added to a button.
+  /// This is an iconData instead of icon so we can control the size
+  final IconData? iconData;
+
   /// Whether the button is enabled or disabled. Buttons are disabled by default. To
   /// enable a button, set its [onPressed] property to a non-null value.
   bool get enabled => onPressed != null;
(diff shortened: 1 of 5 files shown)


skythomp16:
I checked the box about updating documentation but I am just assuming that will get updated because of my comments in the push_button.dart class but is that true? If not, I need to update that an iconData can be passed.

GroovinChip:
> I checked the box about updating documentation but I am just assuming that will get updated because of my comments in the push_button.dart class but is that true? If not, I need to update that an iconData can be passed.

Your assumption is correct.

Would you mind providing a screenshot of what the buttons look like, please?

skythomp16:
Sure, here is a screenshot of the modified example next to a blank canvas in swiftui with a push button with an icon for comparison.
Screenshot 2024-11-01 at 3 02 04 PM

skythomp16:
I pushed an update to fix the tests that were failing. Apologies - I didn't know that I could run 'flutter test' to run all of the tests - I will do that from now on.  I am a little new to Flutter and Dart

GroovinChip:
Do the SwiftUI versions of these buttons have this extra space on the left of the icon, for mini and small control sizes? I think it looks kind of lopsided.
image

skythomp16:
I think you're right.  I pushed a change to allow us to control that padding.  Here is a screenshot of what it looks like now:
Screenshot 2024-11-03 at 4 48 09 PM

skythomp16:
It looks like the dart-code-metrics failing isn't due to anything in my PR

GroovinChip:
Yes, that check needs to be removed

lib/src/buttons/push_button.dart
```diff
@@ -202,6 +231,10 @@ class PushButton extends StatefulWidget {
   /// Can still be overridden if the [color] attribute is non-null.
   final bool? secondary;
 
+  /// An optional iconData that can be added to a button.
+  /// This is an iconData instead of icon so we can control the size
```
GroovinChip, line 235:
```suggestion
  /// An optional `IconData` that can be added to a button.
  /// 
  /// This is an `IconData` instead of an `Icon` so that the icon's size can be controlled internally.
```

lib/src/buttons/push_button.dart
```diff
@@ -361,18 +395,31 @@ class PushButtonState extends State<PushButton>
...
-                            alignment: widget.alignment,
-                            widthFactor: 1.0,
-                            heightFactor: 1.0,
-                            child: DefaultTextStyle(
-                              style: widget.controlSize.textStyle(baseStyle),
-                              child: widget.child,
+                        child: Row(
+                          children: [
+                            widget.iconData != null
+                                ? Padding(padding: widget.controlSize.iconPadding,)
```
GroovinChip, line 401:
```suggestion
                                ? Padding(padding: widget.controlSize.iconPadding)
```

lib/src/buttons/push_button.dart
```diff
@@ -361,18 +395,31 @@ class PushButtonState extends State<PushButton>
...
+                                : const SizedBox.shrink(),
+                            widget.iconData != null ? Icon(widget.iconData,
```
GroovinChip, line 403: Please change this `Icon` to a `MacosIcon`

lib/src/buttons/push_button.dart
```diff
@@ -361,18 +395,31 @@ class PushButtonState extends State<PushButton>
...
+                                  alignment: widget.alignment,
+                                  widthFactor: 1.0,
+                                  heightFactor: 1.0,
+                                  child: Row(children: [
+                                    DefaultTextStyle(
+                                      style: widget.controlSize.textStyle(
+                                          baseStyle),
+                                      child: widget.child,
+                                    ),
+                                  ],)
```
GroovinChip, line 418: Please run `dart format`on this file

GroovinChip (changes requested):

skythomp16:
This PR is old and it looks like I never finished it up. I will close and reopen if I get time to fix these issues.

PR #543, opened 2025-01-07
SidebarItems section header

whirlun:
Add an additional `section` parameter to SidebarItem to convert it into an unclickable widget so it can be used as a section header. It may be better to design a dedicated `Section` widget for this task but this workaround works well for me now.
Capture d’écran, le 2025-01-07 à 3 32 48 p m

## Pre-launch Checklist

### DIFF
--- a/lib/src/sheets/macos_sheet.dart
+++ b/lib/src/sheets/macos_sheet.dart
@@ -114,7 +114,10 @@ Future<T?> showMacosSheet<T>({
   barrierColor ??= MacosDynamicColor.resolve(
     MacosColors.controlBackgroundColor,
     context,
-  ).withValues(alpha: 0.6);
+  );
+
+  barrierColor = Color.fromRGBO((barrierColor.r * 255).floor(),
+      (barrierColor.g * 255).floor(), (barrierColor.b * 255).floor(), 0.6);
 
   return Navigator.of(context, rootNavigator: useRootNavigator).push<T>(
     _MacosSheetRoute<T>(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [2.1.5]
+### 🔄 Updated 🔄
+* add `header` parameter to `SidebarItem` to display an unclickable widget in the sidebar as a section header.
+
 ## [2.1.4]
 ### 🛠️ Fixed 🛠️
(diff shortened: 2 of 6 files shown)


Adrian-Samoticha:
Thanks. Does this support headers that allow the user to fold the content like native headers do:


https://github.com/user-attachments/assets/4b95b4b8-1338-4d67-ad9e-c3528c650c42

And if not, do you think it can be added to your solution easily?

whirlun:
> Thanks. Does this support headers that allow the user to fold the content like native headers do:
> 
>  Screen.Recording.2025-01-08.at.14.44.51.mov

Sure, I've added this functionality by reusing disclosureItems. Now if section is true and disclosureItems is not null it will have the ability to expand.

https://github.com/user-attachments/assets/47871d67-2852-4565-b0ab-560833563040

Adrian-Samoticha:
Wow, that’s great! I’ll review this PR soon.

lib/src/sheets/macos_sheet.dart
```diff
@@ -114,7 +114,10 @@ Future<T?> showMacosSheet<T>({
   barrierColor ??= MacosDynamicColor.resolve(
     MacosColors.controlBackgroundColor,
     context,
-  ).withValues(alpha: 0.6);
+  );
+
+  barrierColor = Color.fromRGBO((barrierColor.r * 255).floor(),
```
tp, line 119: @Adrian-Samoticha Can you please explain why this code would be different from the `withValues(alpha:)` before?

I can't see how this would not result in the same color, but maybe I am missing something.

Gave this a quick spin in DartPad to verify, but I always end up with the same color in either approach 🤔


CleanShot 2025-01-16 at 08 55 47@2x

Adrian-Samoticha: @tp I don’t remember the details, but if I recall correctly, `barrierColor` wasn’t a standard `Color` object in this case, but some object that inherited from `Color` and whose actual color value was resolved at runtime. It was actually correctly resolved (as in, it was resolved to some object that had a property that held the correct value), but was still rendered incorrectly (it was white even in dark mode).

I just forced it to be a standard `Color` object that is easier to reason about to make sure that it produces the correct color regardless of the app’s current light/dark-mode setting. There may have been a better way to go about this, but this was I believe the simplest approach.

PR #559, opened 2025-03-01
MacosTextField : Default cursor color to text color when not specified

driftwoodstudio:
MacosTextField has optional parameter .cursorColor that, if not specified, defaults to simple white or black (#fff, #000) depending on current ThemeMode. 

If the MacosTextField does not _always_ use default theme background color, the default cursor color may not be visible. This requires the dev to code it so if text color is non-default, the cursor color must also be overridden to non-default.

A more sensible fallback for "if not specified, use..." is to match the cursor color to the text. Just blanket forcing black or white is a poor choice when the text color is available to be examined instead.

Example: A writing app, where user can specify text color and background color. The user could be using any text color, on any background color, regardless of whether or no the app itself is in Dark or Light mode. 

This change inserts "check text color" as an option, before giving up and simply choosing black or white with no regard for context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [2.1.10]
+### 🔄 Updated 🔄
+* Make text field cursor color follow the user’s accent color if not explicitly set (thanks, [@driftwoodstudio](https://github.com/driftwoodstudio)).
+* Expose the `AccentColorListener` class to macos_ui users.
+
 ## [2.1.9]
 ### 🛠️ Fixed 🛠️
 * Fix incorrect highlighting and focusing behavior of `_MacosPopupMenuItemButton`.
--- a/lib/macos_ui.dart
+++ b/lib/macos_ui.dart
@@ -90,3 +90,4 @@ export 'src/theme/tooltip_theme.dart';
 export 'src/theme/typography.dart';
 export 'src/enums/accent_color.dart';
 export 'src/utils/window_main_state_listener.dart';
+export 'src/utils/accent_color_listener.dart';
--- a/lib/src/buttons/checkbox.dart
+++ b/lib/src/buttons/checkbox.dart
@@ -365,9 +365,6 @@ class _BoxDecorationBuilder {
             MacosColor.fromRGBO(148, 148, 148, 1.0 * isEnabledFactor),
             MacosColor.fromRGBO(148, 148, 148, 1.0 * isEnabledFactor),
           ];
-
-        default:
-          throw UnimplementedError();
       }
     } else {
       switch (accentColor) {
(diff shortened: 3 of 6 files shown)


Adrian-Samoticha:
To be honest, shouldn’t the color match the current accent color, instead? I believe that’s how macOS does it (most of the time at least, it seems that Launchpad uses a white cursor, but that’s not really a window so it probably doesn’t count).

driftwoodstudio:
If you're asking my opinion/preference, then I'd say No. The accent color is just another arbitrary default that may not be visible at all, or extremely low contrast, depending on what background color the text is on. At least the text color was chosen by the user to be visible on top of the background, so matching that choice by the user seems a more reliable default vs picking one of three arbitrary colors without regard for contrast/context (accent, black, white). 

I don't personally remember seeing accent color used as a cursor color, but there aren't many apps like mine with this sort of "the user gets to pick background and foreground-font colors while editing text" so it's not a common situation.

Adrian-Samoticha:
I took a look around and found these text fields (my accent color is purple):

**Pages:**
image

**Search field in the Preferences app:**
image

**Spotlight Search:**
image

I know it’s a little bit more work, but macos_ui has a class called the `AccentColorListener` whose `currentAccentColor` provides the current accent color as an enum. You’d need to translate that color into a Flutter color yourself.

Perhaps I might push changes to this PR to do that.

**EDIT:** Regarding your point about the contrast; I’d argue that if the user (I assume you’re talking about the Flutter developer that uses macos_ui for their app, and not the user of the app itself) decides to change the background color of the text field to a non-default value, they might as well do the same for the cursor color. It therefore makes more sense to provide sensible defaults (that closely match native behavior) instead.

**EDIT 2:** Wait, my assumption was wrong. I see you’re making a document editing app. Either way, the cursor follows the accent color in Keynote (which allows the user to change the background color) as well:

image

The issue with the low contrast exists, but it’s native behavior. If you wish to change that, just change the cursor color manually, I guess.

driftwoodstudio:
Those are pretty mild examples in terms of "alternate" background color. The background color is just a small shade variant on the canvas background color.

Here's a couple examples from my outliner app, with some (unattractive) style examples applied (all text is editable in-place by clicking on a row).

outliner

outliner 2

For this app, there is no default arbitrary cursor color that can be assumed to work. What will work is entirely dependent on how the user has chosen to style their document.  In all cases, matching the text color would work, because the user choices have (presumably) chosen a text color that works on their chosen background color.

Adrian-Samoticha:
Hmm, that’s an interesting use case. How are you supplying the widgets with the text color, though? Wouldn’t you be able to just provide the cursor color the same way (by just passing it into the `cursorColor` argument)? You’d then be diverging from macOS’ native behavior, but you’d make that decision only for your app, and not for all apps using macos_ui, which I think is preferable.

driftwoodstudio:
I think there's a good argument for changing the default to accent color, as you've suggested.  I think my situation is unusual, and there is a mechanism for addressing it via manual setting of cursor color if needed. 

I've changed my fork of the project to have the behavior I need, and I plan to stick with my fork rather than trying to re-adopt main fork due to the number of small issues I've hit where I'm kind of pushing the boundaries a bit and running into unusual combinations.

PR #573, opened 2025-06-04
Add support for Swift package manager

EchoEllet:
# Swift Package Manager support

This PR is the same as https://github.com/macosui/macos_window_utils.dart/pull/66 but for `macos_ui`.

The PRs https://github.com/Adrian-Samoticha/appkit_ui_element_colors/pull/2 and https://github.com/macosui/macos_window_utils.dart/pull/66, in addition to this one, make changes for `macos_ui` and its plugin dependencies to support the Swift package manager without using CocoaPods.

## Testing

1. Run the following:

  ```bash
  flutter config --enable-swift-package-manager
  ```

2. Add the following to your `pubspec.yaml`:

```yaml
dependency_overrides:
  # Remove the override if this PR is merged: https://github.com/macosui/macos_window_utils.dart/pull/66
  macos_window_utils:
    git:
      url: https://github.com/EchoEllet/macos_window_utils.dart
      ref: feat/swift-package-manager
  # Remove the override if this PR is merged: https://github.com/macosui/macos_ui/pull/573
  macos_ui:
    git:
      url: https://github.com/EchoEllet/macos_ui
      ref: feat/swift-package-manager
```

3. Run `flutter pub get` in your app project, and you should be able to build without CocoaPods integration or CocoaPods installed on your system.
4. Run `flutter pub upgrade` to ensure the transitive dependency [appkit_ui_element_colors](https://pub.dev/packages/appkit_ui_element_colors) is `1.0.1` or newer (to ensure https://github.com/Adrian-Samoticha/appkit_ui_element_colors/pull/2 is applied).

## Related Issues

* *Fix #572*
* *Related https://github.com/Adrian-Samoticha/appkit_ui_element_colors/issues/1*
* *Related https://github.com/macosui/macos_window_utils.dart/issues/52*

## Pre-launch Checklist

- [x] I have incremented the package version as appropriate and updated `CHANGELOG.md` with my changes 
- [ ] I have added/updated relevant documentation 
- [ ] I have run "optimize/organize imports" on all changed files
- [x] I have addressed all analyzer warnings as best I could

PR #580, opened 2025-09-29
chore: Housekeeping

GroovinChip:
This PR takes care of some minor housekeeping:
* Support Dart 3.9.2
* Bump dependencies & add required new direct dependency `vector_math`
* Address lints & migrate away from the deprecated `scale()` function to `scaleByVector3()` & according increase minimum supported Flutter version to match

## Pre-launch Checklist


Adrian-Samoticha:
I get this when trying to run the project with Flutter 3.35.5:

```
All plugins found for macos are Swift Packages, but your project still has CocoaPods integration. Your
project uses a non-standard Podfile and will need to be migrated to Swift Package Manager manually. Some
steps you may need to complete include:
  * In the macos/ directory run "pod deintegrate"
  * Transition any Pod dependencies to Swift Package equivalents. See
  https://developer.apple.com/documentation/xcode/adding-package-dependencies-to-your-app
  * Transition any custom logic
  * Remove the include to "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig" in your
  macos/Flutter/Flutter-Debug.xcconfig
  * Remove the include to "Pods/Target Support Files/Pods-Runner/Pods-Runner.release.xcconfig" in your
  macos/Flutter/Flutter-Release.xcconfig

Removing CocoaPods integration will improve the project's build time.
Launching lib/main.dart on macOS in debug mode...
Warning: CocoaPods is installed but broken. Skipping pod install.
  You appear to have CocoaPods installed but it is not working.
  This can happen if the version of Ruby that CocoaPods was installed with is different from the one being
  used to invoke it.
  This can usually be fixed by re-installing CocoaPods.
For re-installation instructions, see https://guides.cocoapods.org/using/getting-started.html#installation



Error: CocoaPods not installed or not in valid state.
```

This is after enabling SPM with the following command:

```
flutter config --enable-swift-package-manager
```

I would have guessed Cocoapods’ state wouldn’t matter if I’m using SPM. Note that #573 adds support for SPM, so I’m not sure what’s causing the issue. Does the same thing happen on your machine @GroovinChip?

GroovinChip:
This does not occur for me @Adrian-Samoticha

PR #585, opened 2025-10-13
Fixed setState issue in MacosPulldownButton

Abbas1Hussein:
### 🔄 Reopened PR

This replaces a previous PR that was closed due to a branch issue.
Reopening with a clean branch and the same fix applied.

---

### 🧩 Fix: avoid calling setState after dispose in MacosPulldownButton

When I used `MacosPulldownButton`, this error appeared during tests:

```
setState() called after dispose(): _MacosPulldownButtonState (lifecycle state: defunct, not mounted)
```

It happened because `setState` was called inside the `navigator.push(...).then(...)` callback after the widget was already disposed.

I fixed it by adding a `mounted` check before calling `setState`:

```dart
navigator.push(_pulldownRoute!).then<void>((_) {
  if (!mounted) return;
  setState(() => _pullDownButtonState = PulldownButtonState.enabled);
  _removeMacosPulldownRoute();
});
```

Thanks for your patience, I’ll be happy to make any additional changes if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [2.2.2]
+### 🛠 Fixed 🛠
+- Fixed setState called after dispose issue in MacosPulldownButton.
+
 ## [2.2.1]
 ### 🔄 Updated 🔄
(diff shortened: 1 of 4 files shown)


GroovinChip (approved):
LGTM

